PR #42, opened 2012-04-01
Bugfix: use extra headers -- allows Requester Pays

brendano:
for buckets you don't own.

I just used the patch on: http://arxiv.org/help/bulk_data_s3

Here is some discussion about it: http://permalink.gmane.org/gmane.comp.file-systems.s3.s3tools/405


andreabedini:
This patch was reverted by @mdomsch on 19 Jan with commit 77f3a1ac60c32e6828f9a48161f8fb90d7324ba6.

mdomsch:
upstream master has a --requester-pays command line option now for such buckets.

matanster:
Thanks :)

PR #74, opened 2012-08-22
Multipart copy

porcupie:
I have added support for Multipart Copy (s3 to s3) in porcupie s3cmd/multipart-copy branch. In this way one can remotely copy large files (> 5gb). It reuses the existing MultipartUpload functionality by subclassing and passing byte-ranges of the remote file (copy) instead of actual byte-ranged data from a local file (upload).

It might be useful to others.  I see a few branches out there with more interesting Multipart functionality like threaded uploads and more robust error handling but have not tried to integrate with those features yet.


fviard:
8 years later, it is finally merged and fixed in MASTER!
(champagne_smiley)
Thank you @porcupie for the original PR.

PR #96, opened 2012-12-20
Trouble with Windows file paths

zatchery:
I recently had trouble with downloading a WindowsImageBackup file to my Windows Server 2008 machine. After taking it to the debugger I found that python assumed makedirs on line 423 in s3cmd is being passed a valid file path (basename) which is constructed on line 386. The problem is that in windows \ is the path separator (os.path.sep) vs. the / in everything else. I fixed this by using the str.replace("a", "b") function to switch the / with the os.path.sep on line 386 which gets the correct path separator from the OS.

### DIFF
--- a/s3cmd
+++ b/s3cmd
@@ -383,7 +383,7 @@ def cmd_object_get(args):
             if destination_base[-1] != os.path.sep:
                 destination_base += os.path.sep
             for key in remote_list:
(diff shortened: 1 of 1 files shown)


fviard:
Closing obsolete PR.
Code has been fixed for this issue a long time ago.

PR #112, opened 2013-02-12
begin support of json formatted output: start with 'ls' command

ghost:
This patch enhances the 'ls' command to support emitting JSON as output.  We're using this for automation around the use of the s3cmd in scripts.


reignmaker:
@mludvig 
> S3cmd is an end-user tool, not really intended as an engine for more complex operations.

Doesn't this break the concept of being a CLI tool? I was trying to get JSON output and found this "ancient" issue...

PR #113, opened 2013-02-12
cmd_sync_remote2remote crashes on connection error

KarneAsada:
When syncing 2 buckets I got a fatal error.  I found that cmd_sync_remote2remote wasn't handling the S3RequestError - so I added that to the except statement.


fviard:
Closing obsolete PR.
Now there are retries, and S3REquestError is a sub exception of S3Error.

Issue #209, opened 2013-08-07
How to restrict number of files returning by ls command in recursive mode?


vpoluyaktov:
Is any way to restrict number of files returning by ls command? When you list a bucket with --recursive flag ls  command can return billions of files and the command takes minutes for completion or crashes by timeout. I need only 100 first files from the list so having some command line switch to restrict number of files will be very useful.

fviard:
Closing old ticket.
Issue already resolved in current version with option "--limit".

PR #211, opened 2013-08-08
--truncate-response command line option implemented (#209)

vpoluyaktov:
The problem: ls command with --recursive flag may return billions of entries and it takes minutes for completion or crashes with "Timeout" error (see Issue #209 for more details). 
Resolution: I implemented new command line option --truncate-response that truncates a response of ls command by first 1000 entries.


fviard:
Obsolete PR. There is now a --limit argument in master for that purpose.

PR #239, opened 2013-10-22
Force sync to quit with -1 if ignore_failed_copy is set to False

moatazelmasry2:
sync prints an error or warning if an error occured while syncing and continue work. 
I changed it, so that it fails, if ignore_failed_copy is set to False and continue work otherwise

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ testsuite
 /MANIFEST
 /dist
 build/*
+.pydevproject
+.gitignore
+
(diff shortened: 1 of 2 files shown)


mdomsch:
there are a lot of places where we loop over a set of items, do a function to each individually, and perhaps that function succeeds or fails, and we'd want to fail entirely, or continue on fail.  I'd like to see the code re-factored in all these cases, so the inner code is broken out into its own function with meaningful return code, and then the outer loop construct would test the return code and continue_on_error flag to determine whether or not to continue.

fviard:
A colleague of mine is working on a refactor with this PR:
https://github.com/s3tools/s3cmd/pull/580
The rational is to use "--stop-on-error" when you want the backup to be aborted at the first error instead of continuing that is the current default behavior.

fviard:
So, this should be fixed with the following PR that was merged:
https://github.com/s3tools/s3cmd/pull/580
If you can give it a try, I would be great!

PR #241, opened 2013-10-31
to allow to bind to a specific local IP

cinemapub:
I have a dual-homed NAS (2 network cards, connected to 2 different uplinks). I want to be able to 
- upload via the network connection that is the fastest
- run 2 simultaneous instances that each use a separate upload link (so load balancing/double speed)

For the last option I still need to make a script that splits up the uploads to be done in 2 parts and then starts 2 S3cmd instances with different --interface parameters

### DIFF
--- a/S3/ConnMan.py
+++ b/S3/ConnMan.py
@@ -13,19 +13,24 @@ import socket
  
 class http_connection(object):
     def __init__(self, id, hostname, ssl, cfg):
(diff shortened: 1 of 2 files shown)


smaxx1337:
Hello,

is there any update on this? We would really like to use different interfaces / outgoing IPs in s3cmd, just like "ping -I [IP] destination".

PR #308, opened 2014-03-28
adding support for eucalyptus 

jeevanullas:
Hi,

Following patch add support eucalyptus. It adds service_path configuration for folks who use Eucalyptus walrus without DNS and takes care of create_request a bit to make it work for Eucalyptus.

Tested against Eucalyptus 4.0.0 that uses a new object storage gateway with backend riakcs but should work for s3 and walrus as well.

Please let me know if there are any questions. I am happy to test with this patch against S3 to ensure nothing is broken if you have any test cases.

Cheers!
Deependra

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -792,7 +792,7 @@ class S3(object):
         if bucket:
             resource['uri'] = "/" + self.urlencode_string(bucket)
             if object:
-                resource['uri'] = "/" + self.urlencode_string(object)
+                resource['uri'] += "/" + self.urlencode_string(object)
         if extra:
             resource['uri'] += extra
         if self.config.service_path:


epankala:
:+1:

halrogers-claritas:
:+1:

jamshid:
Not sure what extra support is needed. I use s3cmd with an S3-compatible service whose endpoint is https://storage.example.com and s3cmd works fine.

Here is my ~/.s3cfg. A command like `s3cmd ls s3://mybucket` accesses buckets like this: `GET https://storage.example.com/mybucket`.

```
[default]
access_key = XXX
host_base = storage.example.com
host_bucket = storage.example.com
secret_key = secret
signature_v2=True
use_https = True
```

Issue #312, opened 2014-04-14
DreamHost support missing ACLs, bucket expiration rules


mdomsch:
The test suite runs successfully when pointed at DreamHost DreamObjects service, except for:
- Private/Public ACL setting and getting fails.
- setting, getting, deleting bucket expiration rules all fail.

./run-tests.py 1..20 24..42 44..51 58..
succeeds.

fviard:
There is now run-tests-minio.py  to have a test with a smaller subset of s3.
having tests whitelisted/blacklisted per service could be nice but fit badly with the tendency to run s3 servers on each one own domain.
Also sadly the auto test is currently a lot state-full and so it is hard to randomly skip some tests and still have the auto test working as intended and completing.

PR #313, opened 2014-04-18
testsuite: skip tests unsupported on DreamHost (#312)

mdomsch:
DreamHost doesn't support all of the S3 APIs, specifically the ACLs
and object expiry.  So, don't bother running those tests when the
host_base contains objects.dreamhost.com.


fviard:
Closing as this MR is legacy, and something similar to skip specific tests based on host was pushed to master with:
https://github.com/s3tools/s3cmd/commit/902a1c84a0e1044327995cb84e21151d94afc310

PR #315, opened 2014-04-20
Feature/content md5

mdomsch:


mdomsch:
I think I'm going to drop this.  The new signature-v4 code always calculates and sends a sha256 over the object body, which is confirmed at the receiving side.  I don't see a reason to send a Content-MD5 header anymore.

PR #343, opened 2014-06-06
Fixes for sync command

hoelzro:
I've been encountering an error during sync sometimes, so I fixed it!  The error is detailed in b900e00


fviard:
Closing obsolete issue

PR #345, opened 2014-06-09
delete versioned objects in bucket delete

vamitrou:


christensenjairus:
11 years later I'm wishing for this feature

PR #358, opened 2014-07-14
Ftx stdin encrypt

### LINKED ISSUE #351: Streaming from stdin (-) with the encrypt flag (-e) set fails poorly
When trying to upload from stdin using multipart uploading (the only method) when the local encryption flag is set causes s3cmd to only read and encrypt one file part, resulting in an upload of an unusable file to the bucket.  However, there is no error message or warning indicating that it failed.

kmcnellis:
Enable encryption when uploading from stdin using the -e encryption flag.

Uses a new config parameter gpg_stream with the added argument of -a (armor) to specify the command to use.  Armor is necessary to allow the subprocess to work correctly and write an eof that gpg can understand

A subprocess is added before the multipart upload to pipe stdin into and out of gpg.

Additionally, resolve_list was moved to utils to allow its access in multipart uploads

Fixes #351
Replaces #354

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -73,7 +73,8 @@ class Config(object):
     _doc['delete_removed'] = "[sync] Remove remote S3 objects when local file has been deleted"
     delay_updates = False
     gpg_passphrase = ""
-    gpg_command = ""
+    gpg_command = "/usr/bin/gpg"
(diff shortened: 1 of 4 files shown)


kmcnellis:
Obsolete

PR #363, opened 2014-07-29
Support for SSE-C

jheller:
Support for the recently announced Server Side Encryption with a Customer provided key (SEE-C) feature now available on Amazon S3. See the [Amazon SSE-C documentation](http://docs.aws.amazon.com/AmazonS3/latest/dev/ServerSideEncryptionCustomerKeys.html).

See [this wiki page](https://github.com/jheller/s3cmd/wiki) for usage, new options and caveats.


ZVampirEM77:
So, it is still not supported in s3cmd, is it right?

PR #374, opened 2014-08-07
add keyring support

tmaher:
It's always bugged me that `s3cmd` wants to save the secret_key and gpg passphrase to a plaintext file in the user's homedir.  People like sharing their dotfiles, even when they don't fully realize they may contain credentials.  Further, some organizations/employers put home directories on network volumes (e.g. NFS), which exacerbates the problem.

Mac OS X, Windows, and Linux-when-using-GNOME/KDE all provide their own mechanism for secret storage.  For all three, when the user logs in, their password is hashed.  The hash is kept in memory (but not written to disk) for the duration of the login session.  It's then used as a symmetric encryption key.  When a user wants to store a secret with the OS, it's encrypted by the password-derived key and the ciphertext is persisted to disk via OS-specific files.  Fortunately for us, Python's [keyring](https://bitbucket.org/kang/python-keyring-lib) library abstracts away the OS-specific mechanisms.  We get a nice simple key/value store API.

This PR attempts to use keyring if it's present.  If so, secrets are replaced with the string `_KEYRING_` in the config file.  To migrate an in-place file, you can run `s3cmd --configure`.  Secrets passed in via env var (e.g. AWS_SECRET_KEY) or as CLI options will, of course, continue to override the config file.  There are four new config options to control keyring behavior, documented in the code comments.

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -212,7 +212,7 @@ class Config(object):
             self.keyring_get_acct())
         if not blob: return False
         try:
-            secrets = json.loads()
+            secrets = json.loads(blob)
         except Exception, e:
             return False
         for s in self.keyring_secret_names:


romiras:
Can anyone resolve conflicts and merge this pull request?

romiras:
As well as this thread "GPG passphrase stored as plaintext?" in http://sourceforge.net/p/s3tools/discussion/618865/thread/95280c28/ remains unanswered.

rampageservices:
This is a GREAT to have. Hope someone can resolve and merge as was stated in 2015.

rampageservices:
> As well as this thread "GPG passphrase stored as plaintext?" in http://sourceforge.net/p/s3tools/discussion/618865/thread/95280c28/ remains unanswered.

This was a little shocking to me as well. The fact that it showed my passphrase in plain-text. That should never happen.

PR #383, opened 2014-09-02
Return non-zero code from 'put' command if some files failed to upload

dottedmag:
Debian bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=669002


fviard:
So, this should be fixed with the following PR that was merged:
https://github.com/s3tools/s3cmd/pull/580
If you can give it a try, It would be great!

PR #387, opened 2014-09-04
Look for config file in current working directory first

craigatx:
This allows you to use different config files without the -c option by putting them in the working folder.


mdomsch:
I'm not inclined to do this.  There are several other ways to find the config file.  I don't believe looking in $(pwd) should be one of them. Closing.

PR #388, opened 2014-09-09
Fixed issue with mixed path separators with s3cmd get --recursive on Windows

ghost:
s3cmd get fails on Windows when it needs to create a directory. The s3 path is not localized to local path separators, so the s3 portion of the path is not identified as a directory. 

This issue results in error 183, discussed [here](http://sourceforge.net/p/s3tools/mailman/s3tools-bugs/thread/CAP6d-HXG08_5gfLYAUs6SkGx%2BeszC3GQ2wwOnrHJOk4RQ7wtnw@mail.gmail.com/) and [here](http://www.region3dfg.org/comp/AWS/s3/s3cmd).

This change fixes the issue by localizing path separators (as is done elsewhere s3cmd) when creating the local file path.

### DIFF
--- a/s3cmd
+++ b/s3cmd
@@ -465,7 +465,10 @@ def cmd_object_get(args):
             if destination_base[-1] != os.path.sep:
                 destination_base += os.path.sep
             for key in remote_list:
-                remote_list[key]['local_filename'] = destination_base + key
+                local_filename = destination_base + key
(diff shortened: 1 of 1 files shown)


rhulha:
Great patch, this solved using s3cmd under Windows 7 for me with the get command!

PR #399, opened 2014-10-12
Added ability to retry tests

davidjmemmett:
Test "57  Get current expiration setting" fails every now and then, likely due to the change not being propagated quick enough in S3. Thus, I have added the ability for tests to retry when they fail. The default retry limit I have proposed is 5, and I added an exception for the test in question, raising it to 10 attempts, just in case.

As I was testing this locally, and not on EC2, I added a check in S3/Config.py so that it only tries to look for IAM role credentials when running on EC2. You get this error if you run s3cmd without any environment variables set on a non-EC2 instance. I thought it'd be better than getting a Python error relating to a socket timeout when trying to fetch IAM role credentials.

I believe that there may be a bigger issue with IAM credentials overriding the values set in ~/.s3cfg, but I am yet to write up an issue for it.

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -12,6 +12,7 @@ import sys
 import Progress
 from SortedDict import SortedDict
 import httplib
+from socket import timeout
 try:
     import json
 except ImportError, e:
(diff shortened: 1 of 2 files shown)


fviard:
Sorry, it took me a very very long time to reply to your message.
Before, role_config() was called, and inside it, when there was the "timeout" when iam is not available, the exception handler with juste "raise" the exception (timeout error).

In your proposed change (that is not related to the retry tests), you create the "host is ec2" that catch the "timeout exception" to return a "False". And so, in such a case, the code will continue and no exception will be raised like before.

PR #407, opened 2014-11-02
--no-check-md5 now also affects put combined with --continue-put parameter

seidnerj:
Sometimes there's a need to continuously upload a file as it grows, "--continue-put" helps with that. The problem is that in some cases the file ends up being very big and thus calculating it's MD5 hash is very time consuming. In this kind of scenario one might be willing to "risk it" by not checking the MD5.


albarki:
So till now is it possible to use sha256 instead of md5 when uploading/sync files to/from s3 bucket Or not ?

PR #418, opened 2014-11-15
Match S3 websites for CF invalidation correctly

titanous:
Before this patch the first distribution with a custom origin that had a prefix matching the bucket would be chosen.


fviard:
This should have been fixed with the patch that was merged some times ago?
Please reopen if not.

Issue #417, opened 2014-11-13
Bug Report ascii


maycson-fonseca-movile:
[maycson.fonseca@]$ sudo /teste/s3cmd-1.5.0-rc1/s3cmd -c /etc/s3cmd/testecfg sync -p --check-md5 --no-progress "/opt/graphite/storage/whisper/sbs-engine3dc1/HYBRID/CARRIER_1/skiped/[BR]-Só-para-Maiores---Oi/count.wsp.gz" "s3://opt/graphite/storage/whisper/sbs-engine3dc1/HYBRID/CARRIER_1/skiped/[BR]-Só-para-Maiores---Oi/count.wsp.gz"

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
    An unexpected error has occurred.
  Please try reproducing the error using
  the latest s3cmd code from the git master
  branch found at:
    https://github.com/s3tools/s3cmd
  If the error persists, please report the
  following lines (removing any private
  info as necessary) to:
   s3tools-bugs@lists.sourceforge.net

You have encountered a UnicodeEncodeError.  Your environment
variable LANG=en_US.UTF-8 may not specify a Unicode encoding (e.g. UTF-8).
Please set LANG=en_US.UTF-8 or similar in your environment before
invoking s3cmd.

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

Problem: UnicodeEncodeError: 'ascii' codec can't encode character u'\xf3' in position 114: ordinal not in range(128)
S3cmd:   1.5.0-rc1
python:   2.6.6 (r266:84292, Feb 22 2013, 00:00:18) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-3)]
environment LANG=en_US.UTF-8

Traceback (most recent call last):
  File "/movile/s3cmd-1.5.0-rc1/s3cmd", line 2523, in <module>
    rc = main()
  File "/movile/s3cmd-1.5.0-rc1/s3cmd", line 2441, in main
    rc = cmd_func(args)
  File "/movile/s3cmd-1.5.0-rc1/s3cmd", line 1472, in cmd_sync
    return cmd_sync_local2remote(args)
  File "/movile/s3cmd-1.5.0-rc1/s3cmd", line 1449, in cmd_sync_local2remote
    any_child_failed = _single_process(local_list)
[...]

PR #425, opened 2014-11-26
Issue 417: Fix bug when trying to upload a file with non-ascii chars

luciano-sabenca-movile:
Hi. 
This is a solution to the problem reported at https://github.com/s3tools/s3cmd/issues/417.
Here the solution is quite simple, we just normalize the url, removing non-standard characters.


mdomsch:
This is properly fixed with PR #495 instead.  Closing this PR.

PR #456, opened 2015-01-14
Use more unique prefix for s3cmd tests

andrewgaul:
Users may have other buckets with a $USER- prefix.  Ideally s3cmd
tests would only remove the buckets that it created itself after tests
complete.


andrewgaul:
Sorry, the log messages misled me to believe s3cmd tests nuked a wildcard of bucket names.  Reading the source it is obvious that it will only remove $PREFIX-s3cmd-autotest-$tail.

PR #457, opened 2015-01-15
use v4 auth with CloudFront

nidi3:
This is just a hack that works for me.
I don't know Python nor s2cmd good enough to assure that it is fully correct.

### DIFF
--- a/S3/Crypto.py
+++ b/S3/Crypto.py
@@ -74,8 +74,7 @@ def getSignatureKey(key, dateStamp, regionName, serviceName):
     kSigning = sign(kService, 'aws4_request')
     return kSigning
 
-def sign_string_v4(method='GET', host='', canonical_uri='/', params={}, region='us-east-1', cur_headers={}, body=''):
-    service = 's3'
+def sign_string_v4(service, method='GET', host='', canonical_uri='/', params={}, region='us-east-1', cur_headers={}, body=''):
 
     cfg = Config.Config()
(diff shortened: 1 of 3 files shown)


mdomsch:
Thanks for the attempt.  I agree this is a bug; I don't think this is quite the right fix though.  I don't understand why we need to change sign_string_v4().  hard-coded region string clearly isn't right.

S3/Crypto.py
```diff
@@ -100,18 +99,19 @@ def sign_string_v4(method='GET', host='', canonical_uri='/', params={}, region='
...
             continue
-        canonical_headers += header.strip() + ':' + str(cur_headers[header]).strip() + '\n'
-        signed_headers += ';' + header.strip()
+        normalized_key = key.strip().lower()
+        canonical_headers += normalized_key + ':' + value.strip() + '\n'
+        signed_headers += ';' + normalized_key
 
     # sort headers
-    canonical_headers = '\n'.join(sorted(canonical_headers.split())) + '\n'
+    canonical_headers = '\n'.join(sorted(canonical_headers.split('\n'))) + '\n'
```
nidi3: This is to avoid problems with header values that contain spaces

nidi3:
Sorry for the poor quality of the pull request.
Surely, hardcoded region is wrong. 
One issue with sign_string_v4() is header values with spaces. As to why I take lower case of the header keys, I cannot remember :disappointed:

PR #458, opened 2015-01-15
fix S3 wildcard certificate checking

mdomsch:
First, the implementation had a bug, and did not actually disable the
ssl.match_hostname() check like we originally believed, because the
change to the context.check_hostname=False flag occurred too late.

Second, we want to do the hostname checking ourselves, so we can
continue when presented with an S3 wildcard certificate.  We invoke
the SSLContext.match_hostname() function for all connections, and if
it fails, we check if we're actually looking at an S3 wildcard
certificate and communicating with an S3 host. If so, this is
permissible.  If not, we fail the connection immediately.

This should resolve the concern over disabling SSL certificate
hostname checking.

### DIFF
--- a/S3/ConnMan.py
+++ b/S3/ConnMan.py
@@ -52,6 +52,23 @@ class http_connection(object):
 
     @staticmethod
     def match_hostname_aws(cert, hostname, e):
+        """
+        Wildcard matching for *.s3.amazonaws.com and similar per region.
+
+        Per http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html:
+        "We recommend that all bucket names comply with DNS naming conventions."
+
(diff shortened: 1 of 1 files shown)


S3/ConnMan.py
```diff
@@ -51,9 +51,34 @@ def _ssl_context():
         return context
 
     @staticmethod
+    def match_hostname_aws(cert, hostname, e):
+        san = cert.get('subjectAltName', ())
+        for key, value in san:
+            if key == 'DNS':
+                if value.startswith('*.s3') and value.endswith('.amazonaws.com') and hostname.endswith('.amazonaws.com'):
```
djmitche: This is correct, but is going to look mighty suspicious to some security person someday.  A comment linking back to the original issue might help.

PR #459, opened 2015-01-16
add new option --limit-rate to throttle download/upload speed

thesues:
Signed-off-by: Dongmao Zhang <deanraccoon@gmail.com>


thesues:
So now, call time.time() only if  limit-rate were set.

thesues:
@Domsch, How do  you think this.  It will only call time.time when limit-rate is enabled.

mdomsch:
Pulled, thanks!

Issue #223, opened 2013-09-11
s3cmd (1.1.0-beta3) setacl: grant works, revoke doesn't


astrostl:
```
s3cmd setacl s3://mytestbucket/mytestfile --acl-grant=all:example@email.address
```

That grants FULL_CONTROL, which I can see via `s3cmd info`.

```
s3cmd setacl s3://mytestbucket/mytestfile --acl-revoke=all:example@email.address
```

That appears to do nothing, which I can see via `s3cmd info`, and confirm through the AWS Web Console.  Also of slight interest might be that `s3cmd info` yields FULL_CONTROL when modified via s3cmd, but if I make an operation in the AWS Web Console - like manually deleting the grant I just added - it revokes FULL_CONTROL for the original owner, and breaks it into a suite of READ, WRITE, READ_ACP, and WRITE_ACP.

ppe:
--acl-revoke actually does work but you need the user id which is not easily available.

Just submitted a pull request #468 that would add support for display names in --acl-revoke (the name listed in ACL when doing an 'info' command on an object)

This would allow you to more easily revoke ACL grants by checking the display name from 'info' command's output and then use it to revoke the ACL.

fviard:
Closing old ticket.
Issue resolved with the only drawback that emails can't be used for revoke, only by id or display name.

astrostl:
Not to nit-pick, but - doesn't the above PR, which was merged, purport to fix it?

fviard:
no, in the code names are badly chosen, but "name" is the canonical id et "display name" the user name.
The user name is not the user email.

When you set the acl, it is like filing an xml file with all the acl for this file and pushing it.
As a special feature for some regions, amazon is able to do the association of an email with an user name. But when later you ask for the acl list, you just have the user name and id.

When you want to revoke a grant, s3cmd retrieve the current acl "xml file" and just mask in it the grants that you want to revoke. Then it set again the acl with this modified "xml file".
But so, there is no way to know to which entry an email would be associated.
And also revoking is not like sending "revoke" request to amazon that could do an association but just modifying an existing acl info.
So that 's why it is not possible to revoke by email.

PR #468, opened 2015-01-29
Add support for display names in --acl-revoke (#223)

ppe:
Current --acl-revoke only accepts user canonical ids as grantee specification.

'info' command lists users' display names in ACL entries.
Adding support for display names in --acl-revoke makes it easy to revoke
ACLs based on info command output. This alleviates problem outlined in #223


mdomsch:
Pulled, thanks!

PR #488, opened 2015-03-05
If response['headers'] is empty, then don't update dict.

glogiotatidis:
Ceph sometime returns

{'status': 200, 'headers': {}, 'reason': '', 'data': ''}

which chokes s3cmd because headers is an empty dictionary.


glogiotatidis:
See also https://github.com/deis/deis/issues/3002

mdomsch:
We're adding more and more workarounds for faulty 3rd party product implementations of the S3 API.  I merged this patch, but would also prefer to see Ceph store and return headers correctly.  They gave us a 200 (which to be fair, we ignored), and expected to see headers returned.

PR #498, opened 2015-03-13
Ignore etag in the case of a KMS server side encryption

jrick1977:
According to Amazon the etag is incorrect in the case of a KMS signed server side encryption, show it should be ignored.


fviard:
Should be fixed by #611

PR #521, opened 2015-04-03
Added support for listing StorageClass attribute.

BCrookAtRA:
Disabled by default.  Enable with list_storageclass = True in .s3cfg

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -37,7 +37,7 @@ class Config(object):
     send_chunk = 4096
     recv_chunk = 4096
     list_md5 = False
-    list_storageclass = False
+    long_listing = False
     human_readable_sizes = False
     extra_headers = SortedDict(ignore_case = True)
     force = False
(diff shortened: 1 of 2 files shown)


BCrookAtRA:
As requested yesterday on s3tools-general@lists.sourceforge.net, here is my pull request.

mdomsch:
Thanks.  A couple things:

1) let's just call this "cfg.long_listing' so we have some flexibility to add additional content later.
2) can you add a "-l" command line option?  LIke this:
' optparser.add_option("-l", dest="list_storageclass", action="store_true", help="Produce long listing [ls]") '

3) I think the storage class should come after the size, and before the object name.

Thanks,
Matt

BCrookAtRA:
How's it looking now?  (I assumed you'd want dest=long_listing in your optparser example.)

BCrookAtRA:
@mdomsch Is there anything you'd like to see me write differently/better?

mdomsch:
Merged, thanks!

Issue #549, opened 2015-05-14
Suppress error message on SIGPIPE


iceboundflame:
Hi, I often like to pipe the output of s3cmd ls and get - into programs like head or less. After these programs terminate, I get an unsightly page full of exclamation marks, a traceback, and error messages, pushing the desired content off my screen. This goes against the "Rule of Silence" of Unix philosophy. Is there an option to disable this error message?

mdomsch:
Merged #550 thanks.

PR #550, opened 2015-05-14
Fail silently on SIGPIPE

iceboundflame:
If an IOError occurs due to EPIPE, i.e. writing to a closed pipe, we can ignore it instead of printing the error screen. This makes s3cmd friendlier when used with other tools like `head` or `less`.

Addresses https://github.com/s3tools/s3cmd/issues/549

XXX: Might this be masking more serious errors that could manifest as EPIPE?

### DIFF
--- a/s3cmd
+++ b/s3cmd
@@ -2614,6 +2614,11 @@ if __name__ == '__main__':
         sys.exit(EX_BREAK)
 
     except IOError, e:
+        if e.errno == errno.EPIPE:
+            # Fail silently on SIGPIPE. This likely means we wrote to a closed
+            # pipe and user does not care for any more output.
+            sys.exit(EX_IOERR)
+
         report_exception(e)
         sys.exit(EX_IOERR)
 


mdomsch:
We catch and explicitly handle EPIPE (usually with a retry) in the three
places where we're doing network I/O, so this may be safe.

mdomsch:
I'd rather handle this in output() rather than in the global signal catcher.

diff --git a/s3cmd b/s3cmd
index 3d4f930..7719c6a 100755
--- a/s3cmd
+++ b/s3cmd
@@ -46,8 +46,12 @@ from logging import debug, info, warning, error
 from distutils.spawn import find_executable
 
 def output(message):
-    sys.stdout.write(message + "\n")
-    sys.stdout.flush()
+    try:
+        sys.stdout.write(message + "\n")
+        sys.stdout.flush()
+    except IOError, e:
+        if e.errno != errno.EPIPE:
+            raise
 
 def check_args_type(args, type, verbose_type):
     """NOTE: This function looks like to not be used."""

mdomsch:
https://github.com/mdomsch/s3cmd/commit/7c7d0235ec7db9e9bc5ae53a228381faaeef8868

iceboundflame:
@mdomsch By doing that, I think the process will continue to download data and write to the closed stdout even after receiving an error. It'd be better to exit in that case (as most unix programs do).

mdomsch:
Good point.  OK, will merge yours.

PR #557, opened 2015-05-30
Handle AWS_SESSION_TOKEN environment variable

philsnow:
This allows s3cmd to be used with temporary credentials as provided by aws-cli's `aws sts get-session-token` command, which is useful if your API key is required by a policy to use MFA.

### DIFF
--- a/S3/CloudFront.py
+++ b/S3/CloudFront.py
@@ -539,9 +539,10 @@ class CloudFront(object):
         if not headers.has_key("x-amz-date"):
             headers["x-amz-date"] = time.strftime("%a, %d %b %Y %H:%M:%S +0000", time.gmtime())
 
-        if len(self.config.access_token)>0:
+        if self.config.used_role_config:
             self.config.role_refresh()
-            headers['x-amz-security-token']=self.config.access_token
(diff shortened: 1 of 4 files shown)


philsnow:
@mdomsch -- you look like the most recently-prolific committer
@dudyk -- you worked on the role config stuff before

Is there any chance either of you could take a look ?  Thanks!

mdomsch:
if awscli can get the session token, why can't s3cmd?  Why call aws sts get-session-token, store it in the environment, then invoke s3cmd?  At that point, might as well just use awscli for everything...

philsnow:
I'm not saying s3cmd couldn't get session tokens itself.  I would prefer not to re-invent the wheel.

As for using awscli for everything, s3cmd has better performance and I think the 'sync' command has somewhat different semantics, making it not quite a drop-in replacement for s3cmd in existing scripts.

mdomsch:
I'd much rather merge a patch that gets the session token directly.

philsnow:
fair.   closing this.  if anybody else wants this as-is, it's in my fork.

thanks

PR #559, opened 2015-06-02
Invalidate multiple CF distributions when necessary

jmif:
Add support for invalidating multiple CF distributions with  --cf-inval when the bucket is the origin for multiple distributions

### DIFF
--- a/S3/CloudFront.py
+++ b/S3/CloudFront.py
@@ -11,6 +11,7 @@ import time
 import random
 from datetime import datetime
 from logging import debug, info, warning, error
+from sets import Set
 
 try:
     import xml.etree.ElementTree as ET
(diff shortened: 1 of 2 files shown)


mdomsch:
I have never worked with CF.  Have you tested this? How extensively?  Can you add some tests to our test suite for CF?

Thanks,
Matt

mdomsch:
See also #418

jmif:
Quite extensively, use it in our production JS release pipeline every day.  Unfortunately wont have time to add tests to the codebase in the near future but I'll put in on my radar.

fviard:
@jmif Thanks for your contribution, I'm also not very familiar with CF, but the logic of your patch looks good.
There is a single point that I can see in review, you should use the builtin "set" type instead of the imported "Set". As sets.Set is deprecated since python 2.6 (and we have decided that compatibility with previous versions of python is not needed).

jmif:
Updated and did some light testing but haven't won't be able to test extensively for some time.  Thanks for reviewing.

fviard:
Merged. Thank you and sorry for the delay.

PR #560, opened 2015-06-03
iterative du, rm

mikeklaas:
Doesn't pass tests yet because it doesn't obey --exclude-* directives

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -315,7 +315,7 @@ class S3(object):
     def bucket_list_noparse(self, bucket, prefix = None, recursive = None, uri_params = {}):
         if prefix:
             uri_params['prefix'] = self.urlencode_string(prefix)
-        if (not self.config.recursive) and not recursive:
+        if not self.config.recursive and not recursive:
             uri_params['delimiter'] = "/"
         request = self.create_request("BUCKET_LIST", bucket = bucket, **uri_params)
         response = self.send_request(request)
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -279,7 +279,7 @@ class S3(object):
         return response
 
     def bucket_list_streaming(self, bucket, prefix = None, recursive = None, uri_params = {}):
-        """ Generator that produces <is_dir>, <dir or object> pairs of the contents of a specified bucket. """
+        """ Generator that produces <dir_list>, <object_list> pairs of groups of content of a specified bucket. """
         def _list_truncated(data):
             ## <IsTruncated> can either be "true" or "false" or be missing completely
             is_truncated = getTextFromXml(data, ".//IsTruncated") or "false"
@@ -307,10 +307,7 @@ class S3(object):
                     uri_params['marker'] = self.urlencode_string(current_prefixes[-1]["Prefix"])
                 debug("Listing continues after '%s'" % uri_params['marker'])
 
-            for pfix in current_prefixes:
-                yield True, pfix
-            for item in current_list:
-                yield False, item
+            yield current_prefixes, current_list
 
     def bucket_list_noparse(self, bucket, prefix = None, recursive = None, uri_params = {}):
         if prefix:
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -618,7 +618,7 @@ class S3(object):
     def object_batch_delete(self, remote_list):
         """ Batch delete given a remote_list """
         uris = [remote_list[item]['object_uri_str'] for item in remote_list]
-        self.object_batch_delete_uri_strs(self, uris)
+        self.object_batch_delete_uri_strs(uris)
 
     def object_batch_delete_uri_strs(self, uris):
         """ Batch delete given a list of object uris """
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -270,8 +270,9 @@ class S3(object):
     def bucket_list(self, bucket, prefix = None, recursive = None, uri_params = {}):
         item_list = []
         prefixes = []
-        for is_dir, obj in self.bucket_list_streaming(bucket, prefix, recursive, uri_params):
-            (prefixes if is_dir else item_list).append(obj)
+        for dirs, objects in self.bucket_list_streaming(bucket, prefix, recursive, uri_params):
+            item_list.extend(objects)
+            prefixes.extend(dirs)
 
         response = {}
         response['list'] = item_list
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -268,6 +268,18 @@ class S3(object):
         return response
 
     def bucket_list(self, bucket, prefix = None, recursive = None, uri_params = {}):
+        item_list = []
+        prefixes = []
+        for is_dir, obj in self.bucket_list_streaming(bucket, prefix, recursive, uri_params):
+            (prefixes if is_dir else item_list).append(obj)
+
+        response = {}
+        response['list'] = item_list
+        response['common_prefixes'] = prefixes
+        return response
+
+    def bucket_list_streaming(self, bucket, prefix = None, recursive = None, uri_params = {}):
+        """ Generator that produces <is_dir>, <dir or object> pairs of the contents of a specified bucket. """
         def _list_truncated(data):
             ## <IsTruncated> can either be "true" or "false" or be missing completely
             is_truncated = getTextFromXml(data, ".//IsTruncated") or "false"
@@ -279,10 +291,8 @@ class S3(object):
         def _get_common_prefixes(data):
             return getListFromXml(data, "CommonPrefixes")
 
-
         uri_params = uri_params.copy()
         truncated = True
-        list = []
         prefixes = []
 
         while truncated:
@@ -297,17 +307,15 @@ class S3(object):
                     uri_params['marker'] = self.urlencode_string(current_prefixes[-1]["Prefix"])
                 debug("Listing continues after '%s'" % uri_params['marker'])
 
-            list += current_list
-            prefixes += current_prefixes
-
-        response['list'] = list
-        response['common_prefixes'] = prefixes
-        return response
+            for pfix in current_prefixes:
+                yield True, pfix
+            for item in current_list:
+                yield False, item
 
     def bucket_list_noparse(self, bucket, prefix = None, recursive = None, uri_params = {}):
         if prefix:
             uri_params['prefix'] = self.urlencode_string(prefix)
-        if not self.config.recursive and not recursive:
+        if (not self.config.recursive) and not recursive:
             uri_params['delimiter'] = "/"
         request = self.create_request("BUCKET_LIST", bucket = bucket, **uri_params)
         response = self.send_request(request)
@@ -608,6 +616,12 @@ class S3(object):
         return response
 
     def object_batch_delete(self, remote_list):
+        """ Batch delete given a remote_list """
+        uris = [remote_list[item]['object_uri_str'] for item in remote_list]
+        self.object_batch_delete_uri_strs(self, uris)
+
+    def object_batch_delete_uri_strs(self, uris):
+        """ Batch delete given a list of object uris """
         def compose_batch_del_xml(bucket, key_list):
             body = u"<?xml version=\"1.0\" encoding=\"UTF-8\"?><Delete>"
             for key in key_list:
@@ -624,7 +638,7 @@ class S3(object):
             body = encode_to_s3(body)
             return body
 
-        batch = [remote_list[item]['object_uri_str'] for item in remote_list]
+        batch = uris
         if len(batch) == 0:
             raise ValueError("Key list is empty")
         bucket = S3Uri(batch[0]).bucket()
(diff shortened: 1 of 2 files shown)


mdomsch:
I like the approach here.  Will wait to merge till the testsuite passes again.  Please advise when you think it's ready to pull.

mikeklaas:
@mdomsch Updated.  I found running the test suite reliably quite challenging, but I'm confident in this patch.  

The biggest change is to revert to the old deletion code if the user passes an --exclude directive.  The biggest incompatibility in this patch is by deleting iteratively, we aren't sure exactly how many files will be deleted ahead of time, so it isn't always possible to conform to `max_delete` (however, we can look ahead 1000 files, so any setting under than that is covered, and the implementation will never delete more than `max_delete` files)

s3cmd
```diff
@@ -583,12 +594,79 @@ def cmd_object_del(args):
 
         if not recursive:
             rc = subcmd_object_del_uri(uri_str)
-        else:
+        elif Config().exclude:
             rc = subcmd_batch_del(uri_str = uri_str)
+        else:
+            rc = subcmd_batch_del_iterative(uri_str = uri_str)
```
spicavigo: pep8 thingy. whitespace around =

mikeklaas: I was matching local style

mikeklaas:
@fviard @mdomsch I'd appreciate another look if you have time.  Happy to evolve the patch with feedback, if any.

fviard:
Sorry for the delay, and thanks for your work! I will try to have a look at it today or tomorrow ^^.

s3cmd
```diff
@@ -583,12 +594,79 @@ def cmd_object_del(args):
...
         if not rc:
             return rc
     return EX_OK
 
+def subcmd_batch_del_iterative(uri_str = None, bucket = None):
```
fviard: Maybe you can profit of this PR to have the following functions:
subcmd_object_del_uri
subcmd_batch_del
subcmd_batch_del_iterative
Take an "uri" object in input, instead of an "uri_str" (/bucket). I think that this would simplify the code a little.

s3cmd
```diff
@@ -583,12 +594,79 @@ def cmd_object_del(args):
...
+        raise ValueError("Pass only one of uri_str or bucket")
+    if bucket: # bucket specified
+        uri_str = "s3://%s" % bucket
+    s3 = S3(cfg)
+    uri = S3Uri(uri_str)
+    bucket = uri.bucket()
+
+    to_delete = []
+    deleted = [0]
+    deleted_bytes = [0]
```
fviard: It looks like that "deleted" and "deleted_bytes" should be "int"s and not arrays of int with a single entry??

s3cmd
```diff
@@ -583,12 +594,79 @@ def cmd_object_del(args):
...
             return rc
     return EX_OK
 
+def subcmd_batch_del_iterative(uri_str = None, bucket = None):
+    """ Streaming version of batch deletion (doesn't realize whole list in memory before deleting). 
+
+    Differences from subcmd_batch_del:
+      - Obeys cfg.max_delete, but if it is >1000, will delete files until we are sure that at least 
+        that many files are to be deleted
+      - Only prints out individual file names if <1000 files are to be deleted
```
fviard: I'm not comfortable at all with this behavior.
Things should be consistent and not based on an arbitrary limit.
In my opinion, "subcmd_batch_del_iterative" should output things in the same way as "subcmd_batch_del" do: print each files that is removed.

s3cmd
```diff
@@ -583,12 +594,79 @@ def cmd_object_del(args):
...
+            rc = subcmd_batch_del_iterative(uri_str = uri_str)
         if not rc:
             return rc
     return EX_OK
 
+def subcmd_batch_del_iterative(uri_str = None, bucket = None):
+    """ Streaming version of batch deletion (doesn't realize whole list in memory before deleting). 
+
+    Differences from subcmd_batch_del:
+      - Obeys cfg.max_delete, but if it is >1000, will delete files until we are sure that at least
```
fviard: That is not great at all to have such an inconsistency between batch_del and batch_del_iterative.
To sum up:
In one case, the "del" is cancelled if you have too much entries to remove.
In the other one, there is no file deleted if cfg.max_delete < 1000.
But, if > 1000,  a number of files rounded to the last thousand is deleted. It is not even that you have up to cfg.max_delete file deleted. 

I think that it would be a lot better to fallback to "subcmd_batch_del" if cfg.max_delete > 1000.

s3cmd
```diff
@@ -583,12 +594,79 @@ def cmd_object_del(args):
...
+    bucket = uri.bucket()
+
+    to_delete = []
+    deleted = [0]
+    deleted_bytes = [0]
+
+    def _status():
+        total_size, size_coeff = formatSize(deleted_bytes[0], Config().human_readable_sizes)
+        total_size_str = str(total_size) + size_coeff
+        output(u"Deleted %s -> %s from %s" % (str(deleted[0]).ljust(8), total_size_str.ljust(8), uri))
```
fviard: 1) Should be a "debug" message as for batch_del.
2) The formatting is very strange, if i'm not wrong it will looks like that:
"Deleted 15 -> 3Mb from s3://mybucket/..."
??? total -> size ???

s3cmd
```diff
@@ -583,12 +594,79 @@ def cmd_object_del(args):
...
+
+    def _delete():
+        if not cfg.dry_run:
+            response = s3.object_batch_delete_uri_strs([uri.compose_uri(bucket, item['Key']) for item in to_delete])
+        deleted_bytes[0] += sum(int(item["Size"]) for item in to_delete)
+        deleted[0] += len(to_delete)
+        to_delete[:] = []
```
fviard: to_delete = [] ?

s3cmd
mikeklaas: I'll think about that, but I'm not sure it would help.  Both `subcmd_batch_del` and `subcmd_object_del_uri` pass off the `uri_str` to `fetch_remote_list()`, which expects a string not an `S3Uri`.  I'll see if I can come up with something simpler, though.

s3cmd
mikeklaas: They need to be mutable objects to be modifiable in the closure (`_delete()`).  I can add a note if you think it is unclear

s3cmd
mikeklaas: I agree with you that it should not be arbitrary.  The challenge is that I'd like to use (and have used, with this patch) s3cmd to delete millions of objects.  Printing the filenames was rather much, plus it seemed to bog down processing somewhat.

How would you feel if the filenames were printed using `info()` (both in the normal and iterative version), so that they only appear if the user specifies `-v`?

s3cmd
mikeklaas: That's a good suggestion.

s3cmd
mikeklaas: I'll fix the formatting.  How about `info()` for this?  When deleting gobs of files, the number and cumulative size is a lot more useful than individual file names

s3cmd
mikeklaas: Same issue as above.  `to_delete = []` causes `to_delete` to be a local var in `_delete()` and not reference the outer variable.

mikeklaas:
Thanks for the review!  I'll work on your feedback and get back to you soon.

s3cmd
fviard: Ah yes, now it makes sense.
But, after having looked more thoroughly at this,  I think that it would be better to use function arguments and return values instead?
For the to_delete[:], in fact you can move the one in delete, directly in the function body after the call to _delete().
And for the one in "_check_max_delete", I suppose that you can safely remove it as if we are about to return "False". The main function will also return, and so free the to_delete list.

s3cmd
fviard: I understand your point, but what bothers me is to have s3cmd more consistent.
Currently, when you are uploading or syncing, there is the name of each file processed that is displayed. That would be awkward to not have the same behavior for the rm.
Have you tried the "-q" (quiet) option for your use case? Maybe it could be enough to support such an heavy task. And you could still use the exit code of the program to get the success or failure of the operation.

mikeklaas:
@fviard Updated the patch with your feedback.  Now, the output is indistinguishable from batch delete with the default settings (print all filenames deleted and no summary information).  If you pass `-q -v`, only summary info is printed, which I think is a good compromise (thanks, btw, I didn't notice the `-q` option).

After cleaning up the function it feels like pretty understandable, idiomatic python to me.  If you still think it's unclear, let me know and I'll rework it to not use closures.

thanks again!

s3cmd
mikeklaas: FYI updated the format to `Deleted 15 objects (3Mb) from s3://…`

S3/S3.py
```diff
@@ -297,17 +307,15 @@ def _get_common_prefixes(data):
...
+            for pfix in current_prefixes:
+                yield True, pfix
+            for item in current_list:
+                yield False, item
 
     def bucket_list_noparse(self, bucket, prefix = None, recursive = None, uri_params = {}):
         if prefix:
             uri_params['prefix'] = self.urlencode_string(prefix)
-        if not self.config.recursive and not recursive:
+        if (not self.config.recursive) and not recursive:
```
fviard: Is this a typo?

fviard:
Thank you for your last update.
I'm sorry for the delay to complete the review, but i was feeling that there was still an issue but I was too busy to be able to focus on your PR.

I took the time to complete the review today, and here is a summary of the remaining points:
- For the closure, I would really prefer having input/output parameters for "deleted" and "deleted_bytes". I checked the rest of the code, and nowhere such a construction is already used. I'm pretty sure that it will cause confusion later. For the "to_delete", I think that it could be acceptable to keep it like that. So, that will match more the existing code.

- "subcmd_bucket_usage": you force recursive to "True".
I think that it would be better to respect the "recursive" global configuration. Thus, you could have the additional feature to be able to report the size of the objects that are in a single folder (and not accumulate with the size of the objects in sub folders).

- (This one took me some time to notice)
In bucket list streaming, you have a chunk of X entries (~1000) splitted in 2 LISTs: dir list, file list.
Then, for each one, you yield the individual value to "recurse", where you reconstruct a LIST of dir, then, you yield each individual value to reconstruct a LIST of (arbitrary value) 1000 files to delete in "subcmd_batch_del_iterative".

My opinion is that you should have a single "yield" per chunk of X entries in "bucket_list_streaming".
Something like:
yield current_prefixes, current_list

Then, recurse will do:
dirs.extend(current_prefixes)
yield current_prefixes, current_list

Then "subcmd_batch_del_iterative" will be a lot more straightforward, almost will have to do
_delete(current_list)
So, no more a need for a global to_delete, to check its len or remaining values :-D

To be even a little more memory efficient, we can probably have something like a double "dirs" buffer in "recurse", so you don't have to keep the complete list of dirs in the end.
Maybe something lile:
double_dirs = [ [object], [] ]
cur_index = 0
whille double_dirs[0] and double_dirs[1]:
   cur_dirs = double_dirs[cur_index]
  while(cur_dirs):
       ...
       double_dirs[cur_index and 0 or 1].extend(current_prefixes)
      ...
      ...
  double_dirs[cur_index] = []
  cur_index = cur_index and 0 or 1
That should free the list of the previous entries each time we go down of a sub level.

S3/S3.py
mikeklaas: yep, fixed.  thanks!

mikeklaas:
Thanks for the review!  Here is a new patch.

 * The closure has been rewritten to only use input/output parameters

 * For `subcmd_disk_usage`, it doesn't make sense to have `recursive` be configurable.  The current behaviour is to always find the disk usage of everything under the given prefix (`recursive` only changes how the objects are iterated through).  If we changed `du` to only return the size of objects in the current folder, it wouldn't be backward-compatible 

 * I can see your point about the groups of 1000.  But your solution only works because s3 lists files with the same window size as the delete command.  I could see amazon making a change to allow you to list 10,000 objects at once, which would break this approach.  Also, it is generally more natural to consume things one at a time (like how `disk_usage` works)—it's rarer to want to batch it.

I did, though, realize the `recurse()` function wasn't serving a significant purpose, and just inlined the code into the functions.

Let me know what you think.  Thanks!

fviard:
Hi,
Thank you again for your hard work.
I like better like that.
But still, I'm really annoyed for the delete_iterative because, like that, you are really having a list and then creating a copy of this list through the yield. So, you are losing cpu cycles recreating a list and you are pretty sure to have the memory usage of 2 times your chunk. So, if you have 900 objects, you will have 2 lists of 900 objects.

mikeklaas:
@fviard Is this more what you had in mind?  Thanks.

fviard:
Thank you, that is almost good to merge but I think that you have forgotten to also fix "bucket_list" in S3.py in the last patch?

mikeklaas:
Fixed and tested.  Thanks for the catch!

fviard:
Merged, thank you for your sustained contribution!

PR #574, opened 2015-06-21
Use HTTPS by default

koenrh:
I think it's about time that `s3cmd` defaults to HTTPS when communicating with Amazon S3, just like `aws-cli`. I could be wrong, of course, but I don't see any issues by settings this as the default options.


hrchu:
Some S3 setup in private cloud environment may not support HTTPS.

fviard:
Looks like to be a good idea in my opinion.
It is just the default configuration, anyone with a private cloud will tweak his configuration, so could certainly select https or not whether it is needed.
Anyone else has an opinion on this patch?

mdomsch:
+1. Private clouds must change endpoint DNS names anyhow.

PR #578, opened 2015-07-07
Allow to synchronize files with whitespaces at the beginning or at the end of name

ddzialak:
There is also small fix for message about calculating md5 checksum when using option --no-check-md5


fviard:
Thank you for your PR.
The md5 fix looks ok, but I don't really understand what is the usage for the "from0" option?
Why there would be file names with \0 at beginning or end. And why would an user expect that to the point of activating a specific option for that?

ddzialak:
Just imagine there are files with new line char in name (yes, some
filesystems allow such weird names) then you have no option to pass such
name as source to [put] command. Using --from0 option you can. See rsync
documentation (there is such option). AFAIK Character \0 is the only one
forbidden in all filesystems.
7 lip 2015 17:47 "Florent Viard" <notifications@github.com> napisał(a):

fviard:
Ok, thanks for the explanation, now that makes sense.
I was remembering having seen this option somewhere but wasn't able to remember where.
That said, do you really think that it is a case common enough to have to be supported? That it is not in fact the user that should fix his dataset? Have you personally encountered such a case?
(Note, if this option was to be added, my opinion is that its name is not explicit enough)

ddzialak:
If ask me, I would kill anyone who use whitespaces in filenames ;-)
I don't think it is common case, but right now my target is to create tool
that will automatically backup users dataset (I can't modify it). In my
case I have no '\n' in filenames, but I have to copy files with whitespaces
at the end of filename, so current solution (splitting by lines and
stripping it to have filename) is not acceptable. I think that option does
not make 's3cmd' tool worst at any point. Of course decision if it should
be merged belongs to you ;) For my needs I will use forked s3cmd from my
account.
If you want I can create PR just for fixing message.

Thanks&Regards
Darek

fviard:
I have looked deeper at your patches.
That would be better if you can do a specific PR for the md5 so it can be merged directly.

Regarding the "from0" patch, i'm not against the idea, but in your implementation, because of this rarely use case, you read the entire file in memory so it is not very efficient for the common case.

And for the last patch, i don't understand very well the rational. You are like forcing the "recursive" behavior if a directory is put in the file-from file whereas in my opinion the "recursive" flag is there to prevent recurring directories by accident.

ddzialak:
Yes, right I'm reading whole stdin input in obvious way but it was like
that also before my patch.
Please notice there are only filenames on stdin (NOT object contents).
Previously it was read line by line, but no object was transferred before
ALL lines was read and stored in dictionary so basically right now whole
stdin is already loaded to memory anyway.

The last patch does not force "recursive" option. It only ignore case when
source is a directory and there is no --recursive option and --files-from
option is used. This is because --recursive option have no effect because
ONLY existing files are copied from --files-from. Option --files-from cause
any other entries (not existing entry, directory, symlink to directory,
device file and "whatever" where os.path.isfile("whatever") is False) are
always ignored. Unless that behavior is not changed there is no reason to
fail because of lack of meaningless option.

PR #579, opened 2015-07-08
Don't slurp whole file while computing SHA256 checksum

petr-tichy:
When using large --multipart-chunk-size-mb or --disable-multipart, the whole file was read into memory.


fviard:
Thank you for this. It is a nice catch!
I have pushed a simpler version, let me know if you encounter any issue.

PR #582, opened 2015-07-17
Hostname validation

psi29a:
Certification validation is very important, but sometimes we can't help what the hostname of the server is. We can _not_ rely on the network we go over, but we can guarantee that the certs created are ours.

By default we check and validate the certificate and hostname, but now we have the ability to disable the hostname check.

./s3cmd -c s3cmd.haproxy.cfg --ca-cert root.crt --no-validate-hostname ls
2015-07-17 08:10  s3://d2a1da4c-ef42-4ce8-9511-4b4e304c5460

For example, we've already done this for mongo, pymongo and txmongo:
https://jira.mongodb.org/browse/PYTHON-834


fviard:
I close this PR as the same issue is now fixed with another merged PR.
Thanks for your contribution anyway!

mdomsch:
see 2e8254ae47a8977c9e48c457bcc9724ba616e30d in mdomsch/s3cmd bug/ssl-checking branch.

psi29a:
Looks good, thanks! :)

Issue #577, opened 2015-07-02
s3cmd multipart upload should check response body


edwintorok:
s3cmd doesn't check the response body for multipart upload complete requests, although the API is documented to return errors in the body instead of the HTTP code:
["After Amazon S3 begins processing the request, it sends an HTTP response header that specifies a 200 OK response. While processing is in progress, Amazon S3 periodically sends whitespace characters to keep the connection from timing out. Because a request could fail after the initial 200 OK response has been sent, it is important that you check the response body to determine whether the request succeeded."](http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadComplete.html)

This can be reproduced with LibreS3 1.0 for example if you run out of space during the upload (reach volume quota), s3cmd --debug shows the error xml, but shows no warning/error and exits with code 0.
 For LibreS3 1.1 I'm adding a workaround in LibreS3 to check the quota prior to sending the 200 OK, and send an http error code (and s3cmd properly displays the error) but other errors will still be returned as 200 OK + error in the body.

fviard:
Hi,
Thank you for your bug report.
You are right, it is an important bug.
I will try to have a look at how to fix that once I will have some free time.

fviard:
Hi @edwintorok ,
Could it be possible for your to test the latest master against this issue?
A patch was merged that should fix your issue.
Thanks

edwintorok:
@fviard  thanks, I've added some code to LibreS3 to fail on purpose, and the exitcode of s3cmd is good now and there is a message that there was a failure. It could be improved to print the actual error message though:
```
x -> s3://vol1/x  [part 1 of 2, 15MB]
 15728640 of 15728640   100% in    0s    15.76 MB/s  done
x -> s3://vol1/x  [part 2 of 2, 640kB]
 655360 of 655360   100% in    0s     4.03 MB/s  done
ERROR: Upload of 'x' failed too many times. Skipping that file.
```

--debug shows:
```
[...]

PR #583, opened 2015-07-17
Error check on multipart upload

Tim-Anderson:
This addresses https://github.com/s3tools/s3cmd/issues/577

Signed-off-by: Tim Anderson <tsa@biglakesoftware.com>


Tim-Anderson:
1) I agree with and will correct as part of this PR

2) the "NoSuchKey" is an error that comes directly in the send_request() response which does the raise S3Error(). The send_request() is used by all commands. Probably best to just change the logic in the subcmd_cp_mv() to do the partial.

3) This I can do in another PR after reviewing the logic. There is retrying the chunk and if any chunk cannot be recovered bail from that file.

PR #587, opened 2015-07-22
Install instructions for PIP

taz77:
I found PIP instructions were in the issue queue. I thought it would be great if there were also in the Install file.


taz77:
Typo fixed!

PR #589, opened 2015-07-24
move progress_meter default setting down to Config class

jcwilson:
The `cfg.progress_meter` config setting is stomped on by the `isatty()` value. On NT it then complains that the progress meter is not available. This means we get `"Option --progress is not yet supported on MS Windows platform. Assuming --no-progress."` even if we specifically set `--no-progress`

This change makes it so the `isatty()` value is used if neither `--progress` nor `--no-progress` are specified. Otherwise, use the user-provided setting.


fviard:
Merged, good idea, thanks!

PR #590, opened 2015-07-24
Support signed URL content disposition type

andrewgaul:
This allows clients to download with human-friendly names and allows
browsers to properly interpret media like videos.


kahing:
:+1:

fviard:
Thank you for this PR.
The added auto test is very helpful!

PR #591, opened 2015-07-28
Avoid attempting a multipart upload with more than 10000 parts

wiof:
I found the issue described here
http://sourceforge.net/p/s3tools/discussion/618865/thread/3b563258/
and have written a simple patch that avoids the unnecessary upload of a file that will fail later after a lot of wasted time and bandwidth.
PS: It is my first Python code and my first Git pull request. HTH.

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -90,6 +90,7 @@ class Config(object):
     mime_type = ""
     enable_multipart = True
     multipart_chunk_size_mb = 15    # MB
+    multipart_max_chunks = 10000    # Maximum chunks on AWS S3, could be different on other S3-compatible APIs
     # List of checks to be performed for 'sync'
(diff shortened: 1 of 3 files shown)


fviard:
The idea is good, but I'm a little annoyed by the fact that I guess that a lot of the S3 compatible services should have different limits (10k chunks) than the one of amazon s3.

wiof:
I will try to investigate a little further and provide a better proposal.
Other than Google and DreamHost DreamObjects (from the Readme), is there any other popular S3-compatible service I should check?

mdomsch:
radosgw seems to be popular. Verizon Cloud Storage, EMC ECS are two other cloud storage systems with S3 interfaces.  I suspect we just need a config option (that we'd pass in the config file, not command line, honestly).

PR #597, opened 2015-08-06
Add http port parameter for server connection

klingtnet:
I'm using s3cmd as client for an [ceph](http://ceph.com/) server that listents on port 8080 for HTTP connections, so I had to introduce the port setting to the config file as well as to the http connection to get it working.

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -29,6 +29,7 @@ class Config(object):
     access_token = ""
     host_base = "s3.amazonaws.com"
     host_bucket = "%(bucket)s.s3.amazonaws.com"
+    host_port = 80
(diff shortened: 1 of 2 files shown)


fviard:
Hi
In theory, you can already set the port address in the configuration with the host_base and host_bucket uris like for a normal url:
host_base = s3.my-server.com:80
host_bucket = s3.my-server.com:80

klingtnet:
Strange, I could have sworn that it hasn't worked like this yesterday.

fviard:
No problem, maybe you added it only in one of these 2 address for your previous test.
Anyway, thanks for your interest in fixing s3cmd :-)

Issue #594, opened 2015-08-01
s3cmd QUIETLY ignores the names of non-existent files


UnitedMarsupials:
ss3cmd put /nonexistent1 /correct /nonexisten2 s3://foo/bar
will quietly ignore the absent file(s) instead of properly alerting the user.

Something as simple as below would help debug things immensely, but, of course, the message can be more elaborate and/or use error/warning instead of debug:

````diff
--- FileLists.py
+++ FileLists.py    2015-07-31 19:02:32.000000000 -0400
@@ -248,4 +248,5 @@
                 full_name = os.path.join(root, f)
                 if not os.path.isfile(full_name):
+                    debug("Skipping %s" % full_name)
                     continue
                 if os.path.islink(full_name):
````

fviard:
Fixed with https://github.com/s3tools/s3cmd/pull/599

PR #599, opened 2015-08-06
Warn when skipping non-existing files or symbolic links

shigezobriand:
I think it should be warned if skipping non-existing local files.
And it's much the same for symbolic links.

### DIFF
--- a/S3/FileLists.py
+++ b/S3/FileLists.py
@@ -269,9 +269,11 @@ def fetch_local_list(args, is_src = False, recursive = None):
             for f in files:
                 full_name = os.path.join(root, f)
                 if not os.path.isfile(deunicodise(full_name)):
+                    warning(u"Skipping over non-existing file: %s" % full_name)
                     continue
                 if os.path.islink(deunicodise(full_name)):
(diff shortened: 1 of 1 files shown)


fviard:
Thanks for the patch.
That looks true.
Related to issue: https://github.com/s3tools/s3cmd/issues/594

shigezobriand:
Thanks for linking it.

fviard:
Merged, thank you!

panih2o:
Hi Guys, Isn't this breaking the tests [here](https://gist.github.com/panih2o/b009a5ad5d308cd95975) now, as the test is searching for *.png$ on the stdout?

fviard:
@panih2o : Thank you for the notice, indeed you are right, "permission-denied-dir" is also concerned.
I will fix the test suite shortly.

panih2o:
Thanks @fviard

fviard:
@panih2o: The automatic tests should now be fixed in HEAD with:
https://github.com/s3tools/s3cmd/commit/efaa90fd2673ad1ee419d94e1d3caea1f72bd0e1

PR #603, opened 2015-08-17
Bug/testsuite (fixes #602)

### LINKED ISSUE #602: run-tests.py fails in lots of ways against HEAD
Lots of failures in run-tests.py due to changes in return codes from operations (mostly EX_PARTIALs now), and changes to strings being printed.

mdomsch:


fviard:
@mdomsch:
Hum, havn't you forgot to push the change associated with your last comment?

mdomsch:
I still need to fix it from your review comments.

mdomsch:
@fviard please look again.

fviard:
After the previous fix, what do you think about that?
"But, maybe it would be better to have
if cfg.stop_on_error:
return scoreboard.rc()
inside the if e.code == Nosuchkey code to have the correct return code in such a case ?"

On another topic, after trying to run it against master, I noticed that you should have done it against another version of s3cmd, or one of a pull request, because there are "must_find" match issues with the current master.
- You assume that all transfer output are like:
XXXX -> YYYY
but currently it is most of the time:
'XXX' stored as 'YYYY'

I think that is it is good to use '->' instead of 'stored as'.
But that is probably better to keep the single quote around each entity.

fviard:
@mdomsch I can do the pass on the code to change everything to match the syntax '->' but I need your opinion on whether to have the "singe quote" around the file names. My own opinion is that it is better.

mdomsch:
Please do.

Issue #606, opened 2015-08-26
option 'acl_public' is not effective in s3cfg


gooo000:
My s3cmd's version is 1.5.2 and "acl_public = True" is writed in ~/.s3cfg,  but it's not effective when running "s3cmd put file". The object is still private.

mdomsch:
See #677 which should fix this.

PR #607, opened 2015-08-26
supported NoneType options, such as 'acl_public'

gooo000:
fixed the issue #606


mdomsch:
I looked into this once before, and recall changing this to allow NoneType
to be handled the same way was going to be problematic, but I don't
remember the details.  This would need some careful testing.

mdomsch:
See #677 which is a better fix.

PR #609, opened 2015-08-27
report_exception: dedup message and fix indenting

edunham:
For some reason the report_exception function had the 13-line error message
twice. Its body was also indented by 8 spaces, when the rest of the file uses
4.

### DIFF
--- a/s3cmd
+++ b/s3cmd
@@ -2616,7 +2616,7 @@ def main():
     return rc
 
 def report_exception(e, msg=''):
-        sys.stderr.write(u"""
+    alert = u"""
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
     An unexpected error has occurred.
   Please try reproducing the error using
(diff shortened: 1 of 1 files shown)


fviard:
Hi, thanks for the PR,
but the 13 lines error message being printed twice is a intended feature not a bug ^^

edunham:
Yes, I understand that it was supposed to be printed at the start and end of the error. That's why I factored it out into a variable named `alert` and ensured that it would be printed both times.

fviard:
Hi,
You are right, I'm sorry I did look at it too fast.
So the idea is not bad, but it stays a little thing to fix in your patch:
- There is a subtle difference between both messages: "following lines" -> "above lines".
You should add the support for the difference in the alert message.

fviard:
I just checked in a lightly modified version of your PR:
with https://github.com/s3tools/s3cmd/commit/b282366d33658641f6e489fee03c3427d8dfb421
Thank you for the report and the PR anyway!

edunham:
Sorry about dropping the ball on that one, and thanks for getting the fix in!

Issue #612, opened 2015-09-02
Fix python 2.7.9+ certificate check for S3-compatible services


jamshid:
This is related to https://github.com/s3tools/s3cmd/issues/437, hoping you can provide some direction on what needs to be done to get s3cmd working with S3-compatible a https endpoint like https://me.cloud.example.com.

I'm using master but it still doesn't work on my OSX with python 2.7.10.

```
$ git clone https://github.com/s3tools/s3cmd.git
$ cd s3cmd
$ python setup.py build

# Create a file containing each cert, hand edited to remove non-cert output
$ openssl s_client -connect me.cloud.example.com:443 -showcerts </dev/null 2>/dev/null > mycertfile.pem

# List buckets works
$ ./build/scripts-2.7/s3cmd --no-check-certificate --ca-cert ./mycertfile.pem ls 
2015-02-05 07:51  s3://x
2015-02-04 04:15  s3://yum

# List the "x" bucket, weird that it works, right?! I think it's related to being a single character.
$ ./build/scripts-2.7/s3cmd -d --no-check-certificate --ca-cert ./mycertfile.pem ls s3://x
DEBUG: Using ca_certs_file ./mycertfile.pem
DEBUG: Disabling hostname checking
DEBUG: non-proxied HTTPSConnection(me.cloud.example.com)
DEBUG: Unicodising "Bucket name 'x' is too short (min 3 characters)" using UTF-8
DEBUG: format_uri(): /x/?delimiter=/
...
                       DIR   s3://x/repodata/
...

# Listing any other bucket fails :-(
$ ./build/scripts-2.7/s3cmd -d --no-check-certificate --ca-cert ./mycertfile.pem ls s3://yum
WARNING: Retrying failed request: /?delimiter=/ (hostname 'yum.me.cloud.example.com' doesn't match either of '*.cloud.example.com', 'cloud.example.com')
...
```

This is my ~/.s3cfg:
```
[default]
access_key = XXX
host_base = me.cloud.example.com
host_bucket = %(bucket)s.me.cloud.example.com
secret_key = secret
signature_v2=True
use_https = True
```

jamshid:
It's not a general solution, but a side-effect of always returning False from check_bucket_name() is that it avoids the certificate check failure. If you have any pointers maybe I can come up with a usable pull request.

mdomsch:
Please try with this:


diff --git a/S3/ConnMan.py b/S3/ConnMan.py
index 65b1050..c311efe 100644
[...]

PR #616, opened 2015-09-04
Don't check hostnames in SSL certs when --no-check-certificate (#612)

mdomsch:
We were still checking hostnames in SSL certs when
--no-check-certificate is provided.  This caused self-signed
certificates for S3-like services to fail.

Fixes Bug #612

### DIFF
--- a/S3/ConnMan.py
+++ b/S3/ConnMan.py
@@ -74,6 +74,7 @@ class http_connection(object):
         hostname for the *.s3.amazonaws.com wildcard cert, and for the
         region-specific *.s3-[region].amazonaws.com wildcard cert.
         """
+        debug(u'checking SSL subjectAltName against amazonaws.com')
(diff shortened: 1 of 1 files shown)


fviard:
Please have a look at the following PR:
https://github.com/s3tools/s3cmd/pull/582

Excepting the fact the author created another variable instead of using "no-check-certificate".
He set it in another place, I don't know if it is relevant or not.
(and also he fixed the command line help message for this option).

Regarding the "if conn" ..., just in case of unexpected future issue, I think that it could be preferable to have the check to avoid to be confused by an error like the one that got the 2 reporters when trying to investigate this issue ^^

mdomsch:
That was the only place the --no-check-certificate value was used.  This is
effectively a rename. :-(

fviard:
Indeed you are right, sorry ^^
I looked at it too fast.

PR #621, opened 2015-09-13
Bug/ssl checking

mdomsch:
Similar to #582 but does what we want as discussed there.  Also adds validation for cn-north-1 region SSL certificate.

### DIFF
--- a/S3/ConnMan.py
+++ b/S3/ConnMan.py
@@ -94,7 +94,9 @@ class http_connection(object):
         san = cert.get('subjectAltName', ())
         for key, value in san:
             if key == 'DNS':
-                if value.startswith('*.s3') and value.endswith('.amazonaws.com') and self.c.host.endswith('.amazonaws.com'):
+                if value.startswith('*.s3') and \
+                   (value.endswith('.amazonaws.com') and self.c.host.endswith('.amazonaws.com')) or \
+                   (value.endswith('.amazonaws.com.cn') and self.c.host.endswith('.amazonaws.com.cn')):
                     return
         raise e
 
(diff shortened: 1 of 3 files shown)


S3/ConnMan.py
```diff
@@ -9,6 +9,7 @@
 import sys
 import httplib
 import ssl
+from _ssl import CERT_NONE
```
fviard: Why not take CERT_NONE from ssl directly?

S3/ConnMan.py
```diff
@@ -28,11 +29,27 @@ class http_connection(object):
...
+
+    @staticmethod
+    def _ssl_unverified_context(cafile):
+        debug(u'Disabling SSL certificate checking')
+        context = None
+        try:
+            context = ssl._create_unverified_context(cafile=cafile,
+                                                     cert_reqs=CERT_NONE)
+        except AttributeError: # no ssl._create_unverified_context
+            pass
```
fviard: I'm not expert with python and ssl, but why using the "_create_unverified_context()" function, instead of simply using "context.verify_mode = CERT_NONE" after the  ssl.create_default_context(cafile=cafile) ?
If that makes sense, all of that could be factorized in one function i guess.

S3/ConnMan.py
mdomsch: Good catch.  I hadn't realized it was already available as ssl.CERT_NONE.
Branch updated.

fviard:
And what do you think about my other comment?
"I'm not expert with python and ssl, but why using the "_create_unverified_context()" function, instead of simply using "context.verify_mode = CERT_NONE" after the ssl.create_default_context(cafile=cafile) ?
If that makes sense, all of that could be factorized in one function i guess."

S3/ConnMan.py
mdomsch: I tried that but the ssl library won't let you change it after initial
setup. It throws an error. This patch follows the standard library
recommend usage AFAICT.

fviard:
Thanks for the explanation, that looks ok for me.

PR #623, opened 2015-09-14
Bug/testsuite2

mdomsch:
Two fixes for formatting that were preventing the testsuite from running properly when progress is enabled.


fviard:
Nice catch :-)

fviard:
Nice catch :-)

PR #629, opened 2015-09-17
Add support for Infrequent Access

adrianmoisey:
https://aws.amazon.com/blogs/aws/aws-storage-update-new-lower-cost-s3-storage-option-glacier-price-reduction/

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -103,6 +103,7 @@ class Config(object):
     urlencoding_mode = "normal"
     log_target_prefix = ""
     reduced_redundancy = False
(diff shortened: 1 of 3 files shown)


mdomsch:
Thanks Adrian.   Please see my PR https://github.com/s3tools/s3cmd/pull/630
 which does similar, but instead adds a --storage-class option so we don't
have to add specific storage class options each time Amazon makes a change.

adrianmoisey:
Yours looks much nicer :)

fviard:
So, resolved. Thank you for your contribution.

Issue #634, opened 2015-09-21
--continue option does not work


tikonen:
s3cmd 1.5.2 on OS/X

Using `--continue` flag with `get` command downloads the whole file again from beginning and appends it to the existing file resulting a broken file.

PR #635, opened 2015-09-21
Fix for #634. Partial get start position was dropped

tikonen:
The `recv_file` would drop the start_position if it needed to pass the request to 400 or 403 fallback handlers.


mdomsch:
Thanks for investigating and for the patch!

tikonen:
Added same fix on 307 redirect handling

PR #640, opened 2015-09-30
FIX the command error of s3cmd info

qiankunzheng:
If the get_bucket_requester_pays(self, uri) function fails, "s3cmd info s3://buckets" also fails when s3cmd is used in ceph

Signed-off-by: Qiankun Zheng <zheng.qiankun@h3c.com>

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -376,17 +376,23 @@ class S3(object):
         elif location == "EU":
             location = "eu-west-1"
         return location
-
+    
     def get_bucket_requester_pays(self, uri):
         request = self.create_request("BUCKET_LIST", bucket = uri.bucket(), extra = "?requestPayment")
-        response = self.send_request(request)
-        payer = getTextFromXml(response['data'], "Payer")
+        try:
+            response = self.send_request(request)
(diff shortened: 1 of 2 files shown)


fviard:
Hi,
Thank you for having noticed that!
If think that you can remove that check for error 403 in your try:catch and so return None whetever error is encountered.
(For an example, I see the same issue with another service that is returning 501 Not Implemented)

fviard:
Fixed in upstream with a derived version of your PR:
https://github.com/s3tools/s3cmd/commit/bb08c727887d05e4ec50fde9b80eacc22383db7e

Still, thank you very much for your report.

qiankunzheng:
@fviard 
   hi, Do you mind my asking you some questions? Why don't we have permission to get the bucket    which other users have given us read or write permissions by acl policy of S3 information with "s3cmd info" command

fviard:
@qiankunzheng : are you sure that you have given the following permissions: GetBucketLocation and ListBucket?
http://docs.aws.amazon.com/AmazonS3/latest/dev/example-walkthroughs-managing-access-example2.html

PR #667, opened 2015-11-21
Ignore RestoreAlreadyInProgress exception on restore command

kobake:
While restoring recursively, ignore RestoreAlreadyInProgress exception to continue restoring other items.


fviard:
Merged. Thank you

Issue #647, opened 2015-10-20
UnboundLocalError: local variable 'conn' referenced before assignment


carsten-langer:
With latest version 66e96d0707d3d34140158e93f15d560a222cc48a pulled from github today, I receive the error below.
I use: `./s3cmd --config=s3cfg_clanger multipart s3://cdl-test --debug` when the error occurs.
If I use `./s3cmd --config=s3cfg_clanger multipart s3://cdl-test --debug --no-check-hostname`, it works ok.
I use it against our own internal server, from which I know that the SSL certificate validation will fail.
 Thus I suspect that without the `--no-check-hostname` the `conn = ConnMan.get(...` in line 1067 of S3.py would fail, the try block jumps to line 1086 and then in line 1090 tries to access `conn` which is not initialized yet.

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
    An unexpected error has occurred.
  Please try reproducing the error using
  the latest s3cmd code from the git master
  branch found at:
    https://github.com/s3tools/s3cmd
  and have a look at the known issues list:
    https://github.com/s3tools/s3cmd/wiki/Common-known-issues-and-their-solutions
  If the error persists, please report the
  following lines (removing any private
  info as necessary) to:
   s3tools-bugs@lists.sourceforge.net

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

DEBUG: Unicodising './s3cmd' using UTF-8
DEBUG: Unicodising '--config=s3cfg_clanger' using UTF-8
DEBUG: Unicodising 'multipart' using UTF-8
DEBUG: Unicodising 's3://cdl-test' using UTF-8
DEBUG: Unicodising '--debug' using UTF-8
Invoked as: ./s3cmd --config=s3cfg_clanger multipart s3://cdl-test --debug
Problem: UnboundLocalError: local variable 'conn' referenced before assignment
S3cmd:   1.6.0+
python:   2.7.9 (default, Apr  2 2015, 15:33:21)
[GCC 4.9.2]
environment LANG=None

Traceback (most recent call last):
  File "./s3cmd", line 2813, in <module>
    rc = main()
  File "./s3cmd", line 2721, in main
    rc = cmd_func(args)
  File "./s3cmd", line 1826, in cmd_multipart
    response = s3.get_multipart(uri)
  File "/root/s3cmd/S3/S3.py", line 915, in get_multipart
    response = self.send_request(request)
[...]

PR #668, opened 2015-11-24
Disable hostname checks for SSL connections on RHEL 7.2

atodorov:
In RHEL 7.2 python-libs-2.7.5-34.el7.x86_64 introduced the following change:

```
diff -u httplib.py /usr/lib64/python2.7/httplib.py
--- httplib.py	2014-02-11 14:46:38.000000000 +0200
+++ /usr/lib64/python2.7/httplib.py	2015-11-24 16:33:57.106008991 +0200
@@ -215,6 +215,10 @@
 # maximal line length when calling readline().
 _MAXLINE = 65536
 
+# maximum amount of headers accepted
+_MAXHEADERS = 100
+
+
 class HTTPMessage(mimetools.Message):
 
     def addheader(self, key, value):
@@ -271,6 +275,8 @@
         elif self.seekable:
             tell = self.fp.tell
         while True:
+            if len(hlist) > _MAXHEADERS:
+                raise HTTPException("got more than %d headers" % _MAXHEADERS)
             if tell:
                 try:
                     startofline = tell()
@@ -1159,21 +1165,44 @@
 
         def __init__(self, host, port=None, key_file=None, cert_file=None,
                      strict=None, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
-                     source_address=None):
+                     source_address=None, context=None, check_hostname=None):
             HTTPConnection.__init__(self, host, port, strict, timeout,
                                     source_address)
             self.key_file = key_file
             self.cert_file = cert_file
+            if context is None:
+                context = ssl._create_default_https_context()
+            will_verify = context.verify_mode != ssl.CERT_NONE
+            if check_hostname is None:
+                check_hostname = will_verify
+            elif check_hostname and not will_verify:
+                raise ValueError("check_hostname needs a SSL context with "
+                                 "either CERT_OPTIONAL or CERT_REQUIRED")
+            if key_file or cert_file:
+                context.load_cert_chain(cert_file, key_file)
+            self._context = context
+            self._check_hostname = check_hostname
 
         def connect(self):
             "Connect to a host on a given (SSL) port."
 
-            sock = socket.create_connection((self.host, self.port),
-                                            self.timeout, self.source_address)
+            HTTPConnection.connect(self)
+
             if self._tunnel_host:
-                self.sock = sock
-                self._tunnel()
-            self.sock = ssl.wrap_socket(sock, self.key_file, self.cert_file)
+                server_hostname = self._tunnel_host
+            else:
+                server_hostname = self.host
+            sni_hostname = server_hostname if ssl.HAS_SNI else None
+
+            self.sock = self._context.wrap_socket(self.sock,
+                                                  server_hostname=sni_hostname)
+            if not self._context.check_hostname and self._check_hostname:
+                try:
+                    ssl.match_hostname(self.sock.getpeercert(), server_hostname)
+                except Exception:
+                    self.sock.shutdown(socket.SHUT_RDWR)
+                    self.sock.close()
+                    raise
 
     __all__.append("HTTPSConnection")
```

which causes SSL connections to attempt verification of hostname, which in turn fails and causes #647. With this patch both context.check_hostname and check_hostname are set to False which successfully disables this check.

Please build an updated RPM for EPEL once this is included in master. Thanks!

Btw without this patch test 68  List all was failing on RHEL 7.2, now everything is PASS.

### DIFF
--- a/S3/ConnMan.py
+++ b/S3/ConnMan.py
@@ -123,7 +123,12 @@ class http_connection(object):
                 context.check_hostname = False
             conn = httplib.HTTPSConnection(hostname, port, context=context, check_hostname=False)
         except TypeError:
-            conn = httplib.HTTPSConnection(hostname, port)
+            try:
+                # in case check_hostname parameter is not present try again
+                conn = httplib.HTTPSConnection(hostname, port, context=context)
+            except TypeError:
+                # in case even context parameter is not present try one last time
+                conn = httplib.HTTPSConnection(hostname, port)
         return conn
 
     def __init__(self, id, hostname, ssl, cfg):
--- a/S3/ConnMan.py
+++ b/S3/ConnMan.py
@@ -121,7 +121,7 @@ class http_connection(object):
                 # back.  We then run the same check, relaxed for S3's
                 # wildcard certificates.
                 context.check_hostname = False
-            conn = httplib.HTTPSConnection(hostname, port, context=context)
+            conn = httplib.HTTPSConnection(hostname, port, context=context, check_hostname=False)
         except TypeError:
             conn = httplib.HTTPSConnection(hostname, port)
         return conn


mdomsch:
Arrggghhh!  I hate backwards incompatible changes like this.  After this
change, s3cmd running on < RHEL7 will fail because
httplib.HTTPSConnection.__init__()
doesn't take the context or check_hostname arguments.  Odd though - we've
been passing in context= already.  Oh, I see it...

       try:
            context = http_connection._ssl_context()
            # S3's wildcart certificate doesn't work with DNS-style named
buckets.
            if (hostname.endswith('.amazonaws.com') or hostname.endswith('.
amazonaws.com.cn')) and context:
                # this merely delays running the hostname check until
                # after the connection is made and we get control
                # back.  We then run the same check, relaxed for S3's
                # wildcard certificates.
                context.check_hostname = False
            conn = httplib.HTTPSConnection(hostname, port, context=context)
        except TypeError:
            conn = httplib.HTTPSConnection(hostname, port)


If we make the change as this patch suggests, we could get a TypeError for
two different reasons: one where context= is not present (as we catch now),
and one where check_hostname is not present (which we would catch, but
handle improperly).  Which we would get depends on the version of
python-libs installed on the system, and 7.0 >= RHEL < 7.2 we would fail.
:-(

Can you add the corresponding try/except to catch missing check_hostname
also?

Thanks,
Matt



On Tue, Nov 24, 2015 at 8:43 AM, Alexander Todorov <notifications@github.com

atodorov:
I've created another commit for this case although I don't think you need it. The httplib.py diff above suggests that both context and check_hostname are added together. I've not seen the case where only context exists but not check_hostname. Anyway you can apply both commits or only the first one.

atodorov:
Btw I've found a way to inspect what arguments are there in the constructor. From the inspect module:

```
class H:
    def __init__(self, context=None):
        pass

pprint(dict(inspect.getmembers(H.__init__.__func__.__code__)))
```

Look at the 'co_varnames' attribute. It is ('self', 'context').  If you'd like to go this way we can add ifs for context and check_hostname, build the argument list and pass it to the constructor in one statement. Doesn't look so much nicer than the double try-except though.

mdomsch:
double-try with a good comment will be fine thanks.

On Tue, Nov 24, 2015 at 1:42 PM, Alexander Todorov <notifications@github.com

atodorov:
ping, how does the latest look ?

mdomsch:
Almost.  This unconditionally disables all checking, which I don't think is what we want.
See https://github.com/s3tools/s3cmd/tree/atodorov-ssl_wildcard_fix branch with one more patch beyond yours.

mdomsch:
With this patch, it seems to work on RHEL 6.7, RHEL 7.0, RHEL 7.2, and Fedora 23.  Can someone else try other OSs?

mdomsch:
Centos 7.x hasn't caught up with that RHEL 7.2 change yet.  It has:
python-libs-2.7.5-18.el7_1.1.x86_64
which doesn't have either context= or check_hostname=.

Fedora 23 has 
python-libs-2.7.10-8.fc23.x86_64
which has context= but not check_hostname=. 

I find it odd and annoying that Fedora would not have a patch that was already into RHEL 7.2 though.

mdomsch:
The Centos 7 cr repository has the newer python-libs library.  I've tested with and without the fix branch and have seen it fail without, and succeed with, the fix branch.

mdomsch:
http://koji.fedoraproject.org/koji/taskinfo?taskID=12622836 now has an EL7
build with 1.6.1 that includes this fix.

On Sun, Nov 29, 2015 at 7:56 AM, Alexander Todorov <notifications@github.com

PR #684, opened 2015-12-28
build: Do not throw exceptions if no arguments were passed for setup.py

harshavardhana:
Issue:

```
$ python setup.py
  File "setup.py", line 42, in <module>
    if sys.argv[1] == "sdist":
IndexError: list index out of range
```

This can be avoided and should print help without any arguments.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,20 @@
 import sys
 import os
 
+try:
+    import xml.etree.ElementTree
+    print "Using xml.etree.ElementTree for XML processing"
+except ImportError, e:
+    sys.stderr.write(str(e) + "\n")
+    try:
+        import elementtree.ElementTree
+        print "Using elementtree.ElementTree for XML processing"
+    except ImportError, e:
+        sys.stderr.write(str(e) + "\n")
+        sys.stderr.write("Please install ElementTree module from\n")
+        sys.stderr.write("http://effbot.org/zone/element-index.htm\n")
+        sys.exit(1)
+
 from setuptools import setup
 
 import S3.PkgInfo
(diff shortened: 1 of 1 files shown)


fviard:
Thanks for your PR.
The issue is right, but I don't think that the fix is correct.
With the try:except:pass, you will hide any relevant exception.

The correct fix should probably be "if len(argv) > 1 and argv[1]== ...".
If you fix your PR, I will merge it fast :-)

harshavardhana:
> The correct fix should probably be "if len(argv) > 1 and argv[1]== ...".

Yes, let me do that right now.

harshavardhana:
@fviard - patch is updated with your recommendation.

fviard:
Thanks but there is a remaining issue:
You should keep a try catch around the "os.unlink" because otherwise, if the file doesn't already exist there will be a crash ;)

harshavardhana:
> Thanks but there is a remaining issue:
> You should keep a try catch around the "os.unlink" because otherwise, if the file doesn't already exist there will be a crash ;)

Done!

fviard:
hum, I think you did the opposite of what I wanted to mean.
I think that a "try: except: pass" would be better.
The idea is that we want to remove this file, but if it is not already existing, that is not an issue so that should not crash ;-)

harshavardhana:
> hum, I think you did the opposite of what I wanted to mean.
> I think that a "try: except: pass" would be better.
> The idea is that we want to remove this file, but if it is not already existing, that is not an issue so that should not crash ;-)

Ah!

harshavardhana:
Done!

fviard:
Merged. Thank you very much for your reactivity!

PR #691, opened 2016-01-23
Writing messages to STDERR

BackSlasher:
* Removes messages from output when output is STDOUT
* `--quiet` no longer breaks output to STDOUT

### DIFF
--- a/S3/Progress.py
+++ b/S3/Progress.py
@@ -12,11 +12,11 @@ import time
 import Utils
 
 class Progress(object):
-    _stdout = sys.stdout
+    _stdout = sys.stderr
     _last_display = 0
 
     def __init__(self, labels, total_size):
-        self._stdout = sys.stdout
+        self._stdout = sys.stderr
         self.new_file(labels, total_size)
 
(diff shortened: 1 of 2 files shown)


fviard:
Thank you for your PR, but I don't think that it is a good solution.
Standard output message like "progress" ones are not supposed to go on the stderr of the console.
No one will expect to have anything else than errors on the stderr output stream.

BackSlasher:
Thanks for the quick reply.
1. This approach means output to STDOUT is inherently broken
2. Writing progress messages to STDERR is the standard approach AFAIK in Unix.  `curl` and `wget` both write their progress / messages to STDERR (just checked). What standard utility are you familiar with that writes progress messages to STDOUT (esp when it sometimes writes its actual output there)?

BackSlasher:
Other alternatives:

1. re-implement `--quiet` so it doesn't simply close STDOUT
2. avoid printing progress messages if STDOUT is not a tty

BackSlasher:
Abandoning because won't be merged by owner

PR #694, opened 2016-02-01
Add CLI option for specifying access_token

kaustubh-d:
Add CLI option for specifying access_token, currently only .s3cfg config file had this option. 
My use case required us to pass the token as a CLI option along with access, secret key as the temp creds were generated on the fly for use with s3cmd.

It will be nice to have support for this. Can someone please review and get this merged?

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -132,12 +132,12 @@ class Config(object):
     stats = False
 
     ## Creating a singleton
-    def __new__(self, configfile = None, access_key=None, secret_key=None):
+    def __new__(self, configfile = None, access_key=None, secret_key=None, access_token=None):
         if self._instance is None:
             self._instance = object.__new__(self)
         return self._instance
 
-    def __init__(self, configfile = None, access_key=None, secret_key=None):
+    def __init__(self, configfile = None, access_key=None, secret_key=None, access_token=None):
         if configfile:
             try:
                 self.read_config_file(configfile)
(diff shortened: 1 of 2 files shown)


fviard:
Hi, thanks for your proposal, I see the usage but it is a very bad practice to put the credentials / password directly in the command line. Then, your password could be logged and will be visible to whatever do a "ps" or things like that.
Is it too much cost for you to generate a temporary file?

In my case for example, we have a template of the config that we want.
And we generate a temporary file by replacing or concatenating the credentials to it.
Then, call s3cmd with the "-c" option to pass whatever file as a .s3cfg file.
For additional security, you can also create this file in tmpfs ram volume, like for /tmp.

mouse07410:
Temporary files come with their own baggage (as most everything else). 

A better option here would be assigning password to an env variable within the script, and unsettling it after th command has executed. This is how OpenSSL does it. You don't see the password with with ps.

fviard:
Passing the credentials through env variable is already supported by s3cmd:
AWS_ACCESS_KEY
AWS_SECRET_KEY

mouse07410:
> Passing the credentials through env variable is already supported by s3cmd

And for security reasons, it should be sufficient (with recommendation to not set those env vars globally, or for a long time).

kaustubh-d:
Yeah agree, I think using env variables will be better option. I want to avoid using tmp file given that we can use env var or cli option, may be good to have support for both. One of the use cases is also test automation where having cli option flexibility for token will be helpful, since we anyways have secret key as cli option (agree that not recommended for prod setups). Will update the PR for env variable change and share soon. Thanks for the inputs!

fviard:
Closed with the merge of https://github.com/s3tools/s3cmd/pull/701

kaustubh-d:
Thank for getting this in!!

PR #699, opened 2016-02-19
Adding traceback message to ImportError

denisemauldin:
Displaying traceback on an ImportError helps to determine what location s3cmd is pulling libraries from in cases where there are multiple libraries installed accidentally.


fviard:
I could be wrong, but the traceback should already be displayed because of a print L.2881, just before the "if  type(e) == ImportError"?
Or did I miss something?

PR #700, opened 2016-02-19
Change python2 to python

agsha:
Some older debian machines do not have a symlink called python2
Using python makes it more universal


mdomsch:
Newer machines now have python -> python3. How old is your system that has
python 2.6 but not python2 executable?

agsha:
We're running Debian wheezy

mdomsch:
Per
https://www.debian.org/doc/packaging-manuals/python-policy/ch-python.html

Because upstream has started providing it, there will be a symlink for
/usr/bin/python2 for Wheezy and later releases. See PEP-0394
<http://www.python.org/dev/peps/pep-0394/> for details. Packages must be
careful to depend on a sufficient version of python if they make use of
this symlink.
PEP-0394 further states:

   - in preparation for an eventual change in the default version of
   Python, Python 2 only scripts should either be updated to be source
   compatible with Python 3 or else to use python2 in the shebang line.

So we do this (except on RHEL 6, which due to Amazon package differences
compared to actual RHEL, we have to force to python2.6).

We have not fixed s3cmd to be python3 compatible at this time.

So, why doesn't your Wheezy install have a /usr/bin/python2 symlink?

Thanks,
Matt

fviard:
@agsha Do you have anything to add or reply regarding this subject? otherwise we will close the PR.

agsha:
Thanks Matt, Florent,

Looks like something missing from our wheezy package.
Please close the PR

PR #701, opened 2016-02-19
Read access_token from env variable

marcojakob:
There was already a possibility to provide an access_token through the config file. But it should also be possible to read it from the environment variable. See also #694 

Then I had a problem that s3cmd always tried to refresh the IAM role. This isn't possible in my context and resultet in a lot of retries and very long sync times. So I added an option to disable token refreshing. Role refreshing can now be disabled with the CLI option `--no-token-refresh`.

I hope my change fits the s3cmd style and can be reviewed and merged. Thanks.

### DIFF
--- a/S3/CloudFront.py
+++ b/S3/CloudFront.py
@@ -540,8 +540,8 @@ class CloudFront(object):
             headers["x-amz-date"] = time.strftime("%a, %d %b %Y %H:%M:%S +0000", time.gmtime())
 
         if len(self.config.access_token)>0:
-            # Edit: Do not refresh role.
-            # self.config.role_refresh()
+            if self.config.token_refresh:
+                self.config.role_refresh()
             headers['x-amz-security-token']=self.config.access_token
 
         signature = self.sign_request(headers)
--- a/S3/CloudFront.py
+++ b/S3/CloudFront.py
@@ -540,8 +540,7 @@ class CloudFront(object):
             headers["x-amz-date"] = time.strftime("%a, %d %b %Y %H:%M:%S +0000", time.gmtime())
 
         if len(self.config.access_token)>0:
-            if self.config.token_refresh:
-                self.config.role_refresh()
+            self.config.role_refresh()
             headers['x-amz-security-token']=self.config.access_token
 
         signature = self.sign_request(headers)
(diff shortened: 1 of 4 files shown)


fviard:
@marcojakob @kaustubh-d I'm really sorry, I just realised that I was thinking about access credentials and not about the access_token when reading your PR. Now that makes more sense.
But I'm wondering, what is your use cases if you want to use "access tokens" but not IAM?

marcojakob:
My use case: I'm using AWS Lambda. Whenever the lambda function is accessed it already has an access token in the environment variable (it gets it through the IAM role of the lambda function).

Maybe I don't fully understand IAM, but I think since a Lambda function usually only runs for milliseconds or a max of 5 min. there is no need to refresh the token. Is this correct?

fviard:
I'm not expert in Lambda, but based on your explanation, your idea looks like to make sense.
But if this is the typical use case for the "manually" provided access token, maybe it will be better to have an automatic behavior like: don't refresh at all the access_token when it is provided by the user through cmdline, env, or config.
And thus, not adding an additional "--no-refresh" switch.

marcojakob:
Yes, it could make sense to do this "automatic behavior" when an access_token is provided. I just didn't quite understand how s3cmd was handling access_tokens before and didn't want to break someone's code.

How would you do it? Do you want to adjust the PR?

kaustubh-d:
Yeah, we needed a way to set access_token when running the cli. The creds were generated on the fly using getfederationtoken and then used with s3cmd and hence the temp creds set with access token. Will be good to have either the cli or env variable option (preferred env variable at least). Thanks!

fviard:
@marcojakob Sadly I'm a lot busy recently, so if you can adjust the PR, that would help a lot having it merged fast.
In the end, my opinion of the best way to do that is to have your "token_refresh" variable be private: "_access_token_refresh".
The overall patch is good, just remove the "optparser switches".
And inside the Config class __init__, detect if access_token was already set, and in that case set _access_token_refresh to false.
You can probably do the check for this variable inside the "role_refresh" function to not have the check for it in 2 places.

And by the way, regarding the fact that there is already the secret_key that can be set on the command line, that would be good if you can merge the PR #694 of @kaustubh-d with yours to have all of that in one.

Thank you very much!

marcojakob:
@fviard I've tried to make your suggested fixes and also added the changes from PR #694. Please review.

fviard:
Merged, thank you very much!
FYI, I will just remove the entry when generating config files with run_configure as I don't think it is useful to a lot of people.

PR #702, opened 2016-02-20
Do not close stdout, it might be used when file is downloaded

artms:
If you try downloading file with -q and destination is -, you'll receive such errors:
```
s3cmd -q get s3://some-bucket/somefile - > /dev/null
WARNING: Retrying failed request: somefile (I/O operation on closed file)
```

I do not think it is actually very good solution, actually all logging should be done through some logging wrapper, this way there won't be a need to mangle with stdout...

### DIFF
--- a/s3cmd
+++ b/s3cmd
@@ -2639,7 +2639,6 @@ def main():
     if options.quiet:
         try:
             f = open("/dev/null", "w")
-            sys.stdout.close()
             sys.stdout = f
         except IOError:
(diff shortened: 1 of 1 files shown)


mdomsch:
Ah, I hadn't considered the download to stdout case.  That's valid.

artms:
So will this pull be merged?

fviard:
Merged. Thank you very much for this patch and sorry for the long delay to have it merged.

PR #704, opened 2016-02-20
archive.org

h4ck3rm1k3:
archive.org allows upper case letters and _ in the buckets.

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -1125,7 +1125,7 @@ class S3(object):
             redir_bucket = getTextFromXml(response['data'], ".//Bucket")
             redir_hostname = getTextFromXml(response['data'], ".//Endpoint")
             self.set_hostname(redir_bucket, redir_hostname)
-            info("Redirected to: %s" % (redir_hostname))
+            info("Redirected 1 to: %s" % (redir_hostname))
             return self.send_request(request)
 
         if response["status"] >= 500:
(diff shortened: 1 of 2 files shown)


fviard:
@h4ck3rm1k3 Hum, i'm not sure to understand your patch. Do you mean that you have an error or warning when using s3cmd with archive.org?
The amazon S3 rule is no upper case or _ in bucket names, otherwise it switches on "path based" buckets. A lot of S3 clones, like archive.org only supports path based buckets and thus having Upper case and _ in bucket names.
But, so it should already work for you even if a warning is displayed, no?

h4ck3rm1k3:
this is the patch I needed to make to get it to work.

fviard:
It looks like that it is supposed to be working when having the right configuration:
http://askubuntu.com/questions/32763/script-to-upload-to-internet-archive-archive-org
http://sdf.org/?tutorials/s3cmd_archive

if it is still not working without your patch, can you show us your configuration file (don't forget to scramble the access and secret keys).

h4ck3rm1k3:
I dont have time for this now, you can close the pr if you like

On Mon, Mar 14, 2016 at 8:56 PM, Florent Viard <notifications@github.com>
wrote:

> It looks like that it is supposed to be working when having the right
> configuration:
>
> http://askubuntu.com/questions/32763/script-to-upload-to-internet-archive-archive-org
> http://sdf.org/?tutorials/s3cmd_archive
>
> if it is still not working without your patch, can you show us your
> configuration file (don't forget to scramble the access and secret keys).
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/s3tools/s3cmd/pull/704#issuecomment-196589788
>



-- 
James Michael DuPont
Kansas Linux Fest http://kansaslinuxfest.us
Free/Libre Open Source and Open Knowledge Association of Kansas
http://openkansas.us
Member of Free Libre Open Source Software Kosova http://www.flossk.org
Saving Wikipedia(tm) articles from deletion http://SpeedyDeletion.wikia.com

PR #705, opened 2016-02-20
traceback

h4ck3rm1k3:
use existing error handling for ssl errors


fviard:
The first change doesn't look to be useful anymore.
And for the second one, isn't the error explicit enough? I guess that there is no need for a complete error report with a traceback. No?
Don't hesitate to re-open if you disagree with this.

PR #707, opened 2016-02-24
Added marker support to CloudFront.py to retrieve the next page of di…

weirdall:
Added marker support to CloudFront.py to retrieve the next page of distributions for cflist. Will show the current and next marker as well as maximum items and item count before displaying the cloudfront list. This is concerning the s3cmd cflist command.

### DIFF
--- a/S3/CloudFront.py
+++ b/S3/CloudFront.py
@@ -91,6 +91,9 @@ class DistributionList(object):
         self.info['IsTruncated'] = (self.info['IsTruncated'].lower() == "true")
 
         self.dist_summs = []
+        self.marker = self.info.get('Marker')
+        self.next_marker = self.info.get('NextMarker')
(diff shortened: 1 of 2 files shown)


mdomsch:
Can you do this without the user having to specify a marker (e.g. just make
it automatic) please?

fviard:
@weirdall As mdomsch said, it would be better to have the list cloudfront command list all the entries automatically by following all the markers. Without user interaction. Like what is done for S3 list entries.

zakalibit:
It still makes sense to have a marker option, so we can tell where the listing should start, see https://github.com/s3tools/s3cmd/pull/1211, we have huge buckets in production and we had to list the whole bucket at some point, a long running s3cmd ls can be killed, for some reason, like out of memory, as current impl gather the whole list in memory before printing, so having the marker option allows to resume the ls command.

PR #710, opened 2016-03-02
FIX: update s3cmd.1 --disable man entry option

Shamanoid:
--disable have wrong description cc --enable

PR to fix man in 1.x versions of s3cmd


fviard:
You are right, thank you for the PR.

PR #393, opened 2014-09-23
Add getlifecycle subcommand to see raw bucket lifecycle

kuenishi:

PR #715, opened 2016-03-07
Add Transition to info cmd and add getlifecycle for full xml lifecycle

kurtdavis:
Added 'Transistion' line to 'info' cmd and created new 'getlifecycle' cmd for full xml output of lifecycle.
Example output for 'info' cmd:

```
$> s3cmd info s3://sample.bucket.backups
s3://sample.bucket.backups/ (bucket):
   Location:  us-east-1
   Payer:     BucketOwner
   Transition Rule: all objects in this bucket will transition to 'GLACIER' in '8' day(s) after creation
   Expiration Rule: all objects in this bucket will expire in '373' day(s) after creation
   policy:    none
   cors:      none
   ACL:      sampleUser: FULL_CONTROL
```

based on orig work of [kuenishi](https://github.com/kuenishi): [pull req. 393](https://github.com/s3tools/s3cmd/pull/393)

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -445,6 +445,23 @@ class S3(object):
 
         return response
 
+    def transition_info(self, uri, bucket_location = None):
+        bucket = uri.bucket()
+
+        request = self.create_request("BUCKET_LIST", bucket = bucket, extra="?lifecycle")
+        try:
+            response = self.send_request(request)
+            response['prefix'] = getTextFromXml(response['data'], ".//Rule//Prefix")
+            response['date'] = getTextFromXml(response['data'], ".//Rule//Transition//Date")
+            response['days'] = getTextFromXml(response['data'], ".//Rule//Transition//Days")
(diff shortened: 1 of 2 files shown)


fviard:
@kurtdavis Thank you for trying to fix this patch, but there is still some issues with this patch:
- "info cmd" : There could be multiple expiration or transition rules. Here only the first ones are displayed. It would probably be better to have a single line like "lifecycle rule: 3 expirations, 2 transitions"
- There should not be a "200: Got Lifecycle Policy" text in the "getlifecycle" output, otherwise that prevents to easily use the output.
- I don't know how, but having a way to pretty print the lifecycle rules with a "get" could be nice but is just a bonus.
- There is not only Transition and Expiration rules, but also: NoncurrentVersionTransition and NoncurrentVersionExpiration
That would be awesome if you would like to complete this patch.

fviard:
@kurtdavis Would be great if you can find the time to fix this PR ^^

fviard:
Partially fixed with the merge of:
https://github.com/s3tools/s3cmd/pull/763

PR #720, opened 2016-03-18
Missing etag in HTTP headers, handle gracefully.

harshavardhana:
If 'etag' is missing do not crash 's3cmd',  throw a warning and handle it gracefully.

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -1304,7 +1304,8 @@ class S3(object):
 
         debug("MD5 sums: computed=%s, received=%s" % (md5_computed, response["headers"]["etag"]))
         ## when using KMS encryption, MD5 etag value will not match
-        if (response["headers"]["etag"].strip('"\'') != md5_hash.hexdigest()) and response["headers"].get("x-amz-server-side-encryption") != 'aws:kms':
+        md5_from_s3 = response["headers"].get("etag", "").strip('"\'')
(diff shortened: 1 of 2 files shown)


fviard:
Thank you for your proposal.
I think that it is more a server issue than a client issue but anyway safeguards are not bad.
I think you can have a nicer / shorter patch by just replacing : info['headers']['etag'] by info['headers'].get('etag', '').
If you can rework your PR, I will then merge it.

harshavardhana:
> Thank you for your proposal.
> I think that it is more a server issue than a client issue but anyway safeguards are not bad.
> I think you can have a nicer / shorter patch by just replacing : info['headers']['etag'] by info['headers'].get('etag', '').
> If you can rework your PR, I will then merge it.

Sure, i thought 'try, except' - let me rework it quickly.

harshavardhana:
@fviard - updated.

fviard:
Merged thanks!

mishak87:
When will it be released?

fviard:
@mishak87 it is already in Master. Usually you can use git master version because usually we try to keep it stable.
There is not a release planned yet. There is still a few things that will have to be fixed before. But probably we could have one in not too long.

mishak87:
@fviard Thanks! Today I learned that it is possible to install commits via pip :-)
```
pip install git+https://github.com/s3tools/s3cmd.git@4f07fea9df2fe5b0ed7cfacb50be4291a611bb6f
```

Issue #672, opened 2015-12-04
with --follow-links, the target of symlinks should have the mime-type of the file they point to, not of the symlink


MLopez-Ibanez:
If I have a symlink "index" pointing to "index.html", the mime-type of index should be the one of index.html when using --follow-links.

MLopez-Ibanez:
This commit fixes it: https://github.com/MLopez-Ibanez/s3cmd/commit/a211864b8daf379d8b40ed96c63e13f86c03965d

fviard:
Thanks for your report, but I'm a little surprised by the fix that you propose.
What does currently mime_magic when you pass a symlink?
I guess that it use the filename to determine the type based on the extension, so in your case you don't see anything.

But in that case, I'm a little worried of possible side effects with your patch.
Lets suppose that I have "index.html" that is a symlink to my-index.txt.
With your page the mime type would be determined based on ".txt" instead of ".html" as currently.

(Note: that you can also manually set a mime type for a specific file like the "index" in your case.)

MLopez-Ibanez:
What I'm trying to do is to have clean links (that is, links without .html).

I'm not sure why you would have a html file with extension .txt

However, there are cases when you may have a file without extension and you want to keep this link working even if the file type of the target of the symlink changes.

MLopez-Ibanez:
Others like me have struggled to do this in S3:

http://stackoverflow.com/questions/23463679/s3-static-pages-without-html-extension

http://serverfault.com/questions/650704/url-rewriting-in-amazon-s3

URL rewriting simply doesn't scale and it cannot be easily automatized.

This is something so easy in Apache:

http://stackoverflow.com/questions/2836400/clean-urls-for-static-html-files-with-htaccess-and-mod-rewrite

fviard:
Should be resolved in Master with:
https://github.com/s3tools/s3cmd/commit/d7b60b894b1d58d9dd66ea603d3438f0e576589b
I hope that the change in behavior will not have bad side effects for some persons.
But what you proposed looks indeed to be the most logic behavior.

PR #729, opened 2016-04-20
Fix Issue #672

MLopez-Ibanez:
With --follow-links, the target of symlinks should have the mime-type of the file they point to, not of the symlink. Dereference the link before guessing its mime-type.

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -500,6 +500,10 @@ class S3(object):
             raise ParameterError("You must specify --mime-type or --default-mime-type for files uploaded from stdin.")
 
         if self.config.guess_mime_type:
+            if self.config.follow_symlinks:
+                if os.path.islink(deunicodise(filename)):
+                    filename = os.path.realpath(deunicodise(filename))
+
             if self.config.use_mime_magic:
                 (content_type, content_charset) = mime_magic(filename)
             else:


fviard:
Thanks for your PR, the idea is interesting.
Still I have 2 remarks:
1) Is it possible for your do a PR with a rebase instead of with a "merge" to have a clean history.
2) Is there any reason you think of to do the following test:
   if os.path.islink(deunicodise(filename)):
instead of just using "realpath" for all filenames?

MLopez-Ibanez:
1) How do I do that from the command-line?
2) Not sure why I did it this. Perhaps I did not want to change the filename at all if it was not a symlink. But your proposal seems to work fine for my purposes.

fviard:
Sorry for my reply after a very long time.
In your case, as the patch is short, it would probably be better to create a new clean branch from master and reply your patch on it.
Otherwise, for another future patch, use the following command to rebase based on master:
 git rebase upstream/master
(If we assume that upstream is the github s3cmd master)

fviard:
I did a patch based on your proposal, and it is now in master:
https://github.com/s3tools/s3cmd/commit/d7b60b894b1d58d9dd66ea603d3438f0e576589b

Please give it a try and thank you for the initial PR.

PR #733, opened 2016-04-26
add delete more than two objects of a bucket by send one cmd

zhouruisong:
delete more than two objects of a bucket by send one cmd，this call batch delete method.

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -647,6 +647,16 @@ class S3(object):
         response = self.recv_file(request, stream, labels, start_position)
         return response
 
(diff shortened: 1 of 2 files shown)


mdomsch:
The 'del' command should be updated to do this, rather than add a new command.  Batch deletes do happen in other cases.

fviard:
@zhouruisong I think that you should have done a "feature branch" with your previous pull request, as it is directing to your master branch, I think that you did override your previous change and now this PR is not linking to the commit associated with this PR.

PR #747, opened 2016-06-04
Catch 'connection reset by peer' (errno 104) and retry

mdomsch:
S3 occasionally closes our connection unexpectedly.  We weren't
catching this, but would crash with a traceback when it happens.

Explicitly catch and retry on ECONNRESET.

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -1100,7 +1100,7 @@ class S3(object):
         except CertificateError:
             raise
         except (IOError, Exception), e:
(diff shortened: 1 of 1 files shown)


mdomsch:
Feedback from initial reporter is that this is working as expected.

From: Sheila Reynolds

I just wanted to let you know that my script has been running smoothly for the past 6 hours and has downloaded about 1300 files from S3.

Only once have I seen this error and it seems like it was handled gracefully:

PR #779, opened 2016-08-29
Change proxy SSL URI to relative

cybojanek:
Proxy SSL/HTTPS requests use regular relative URIs, while only HTTP requests use absolute URIs


fviard:
Thank you @cybojanek.
Based on the http specification, in theory the http server should be able to handle relative or absolute URI. So I think that the server that you are using is not really compliant.

But looking at what is commonly done in the wide, it looks like that the consensus is to use relative URI in case of ssl (ex. curl).
So I would merge your PR.

PR #781, opened 2016-08-30
Support for custom S3 api path

kwarunek:
This changes allows user to set path to S3 api instead of default '/'. It will enable s3cmd to work with s3ninja or custom s3-ceph endpoints.

### DIFF
--- a/S3/ConnMan.py
+++ b/S3/ConnMan.py
@@ -161,6 +161,11 @@ class http_connection(object):
         parsed_hostname = urlparse('https://' + hostname)
         self.hostname = parsed_hostname.hostname
         self.port = parsed_hostname.port
+        if parsed_hostname.path and parsed_hostname.path != '/':
+            self.path = parsed_hostname.path.rstrip('/')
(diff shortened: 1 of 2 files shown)


fviard:
@kwarunek Thank you for your proposal, it is nice to see that we can be compatible with a custom api_endpoint with a so small change.

But if you are willing to rework your patch, I think that it will be greater to support "api_endpoints" not by having a custom config key, but just by adding the api_endpoint to host_base / host_bucket.
Ex.: host_base = s3.myserver.com:2233/custom_endpoint_path/blurp/
Currently, we already deal with all the stripping necessary in "ConnMan".
There is probably not so much function to fix to support that.

kwarunek:
ok, I'll work on it

fviard:
That looks great for me.
Would you mind doing a rebase/squash to have a single commit based on master? otherwise I will have to manually merge it as there is a conflict.

fviard:
Merged, thank you very much!

kwarunek:
awesome, thanks

PR #784, opened 2016-09-02
Enable to configure "host_base" as "S3 Endpoint" in

mozawa:
Currently, if user wants to use an S3 endpoint rather than Amazon S3 default region endpoint, the following entries need to be updated manually after the first configuration using "--configure".
This change is to enable to configure "host_base" as "S3 endpoint" in the first interactive configuration and update "host_bucket" as well so that user can use this command without any changes after the first configuration.
 
[.s3cfg]
host_base = s3.amazonaws.com
host_bucket = %(bucket)s.s3.amazonaws.com

### DIFF
--- a/s3cmd
+++ b/s3cmd
@@ -2160,6 +2160,8 @@ def run_configure(config_file, args):
         ("access_key", "Access Key", "Access key and Secret key are your identifiers for Amazon S3. Leave them empty for using the env variables."),
         ("secret_key", "Secret Key"),
         ("bucket_location", "Default Region"),
+        ("host_base", "S3 Endpoint", "Use \"s3.amazonaws.com\" for S3 Endpoint and not modify it to the target Amazon S3."),
+        ("host_bucket", "DNS-style bucket+hostname:port template for accessing a bucket", "Use \"%(bucket)s.s3.amazonaws.com\" to the target Amazon S3. \"%(bucket)s\" and \"%(location)s\" vars can be used\nif the target S3 system supports dns based buckets."),
         ("gpg_passphrase", "Encryption password", "Encryption password is used to protect your files from reading\nby unauthorized persons while in transfer to S3"),
         ("gpg_command", "Path to GPG program"),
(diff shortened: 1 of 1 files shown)


mozawa:
@fviard Is it possible to get this feature into the next official release?

fviard:
Hi, sorry for the delay, i wanted to comment but lacked the time.
The base idea is good I think, but there is 2 points i think:
- [Minor] Probably it will be useful to have the default value to use to target amazon in a description. And to suggest not to modify it for using with aws.

- [Major] I'm annoyed because I think that you can't "auto set" the host_bucket base on the "host_value". Because most of custom s3 servers don't use dns based buckets but path style buckets. But people could also be lost in the opposite case if you just  set host_bucket=host_base.
So, in my opinion, it would probably be better to have 2 configuration questions. One for each. And I think that it is important to give the default value for aws s3 and to indicate the "%(bucket)s" and "%(location)s" vars that can be used.

mozawa:
@fviard Thank you for the advice. I made some changes based on your suggestion. Please review this again. Thanks!

fviard:
Thank you for being so reactive.
- Sorry, but I was not thinking about having the "%(bucket)s" automatically set but more that it should have been in the help doc. I think that automatically set it is bad, as most custom configs don't use it.
- Also, by default you will see amazon url in the "helper", but in fact it shows the value that is already in .s3cfg or otherwise the default value. So if you save the config, and run configure a second time, it is your custom value that will be the default. That's why I'm thinking that it could be good to have the amazon default value in the help doc string also.

- Finally, for the website_endpoint that change based on use_https, that could have been a good idea for resolving the case of a few people that might use that. But sadly i don't think that we can do that, because by default aws s3 will not support HTTPS traffic for website_endpoint except if your bucket name is correct and you do a complex setup.
- Also, I think that it is not a common feature, so maybe it is better to not put website_endpoint in the config to not confuse new comers?

mozawa:
@fviard Thank you for reviewing this again. Made changes which got simpler. Thanks!

fviard:
That is perfect enough like that, thank you!

PR #785, opened 2016-09-09
Python3: use print function

shaform:
A small patch.


fviard:
A perfect start, thank you! Merged.

PR #787, opened 2016-09-11
fix: handle empty md5 in find_md5_one

Abhishekvrshny:
`s3cmd sync` was found to be misbehaving with ceph v0.94.2. It was due to a bug in radosgw that returned empty ETags. http://tracker.ceph.com/issues/12955. This PR handles such cases with empty ETags.


Abhishekvrshny:
@fviard @shaform : Can you please review this?

fviard:
Merged. thank you!

PR #789, opened 2016-09-14
Python 3: Fix has_key, except, numliterals with 2to3

shaform:
Some additional small fixes.


fviard:
Merged, thank you.
In my opinion, the code is even more readable with this patch ^^

PR #793, opened 2016-10-11
Make 'acl_public' settings to honor false values

hrubi:
Explicitly treat the 'acl_public' option in configuration file as a
boolean.

As the 'acl_public' option defaults to None, it wasn't treated as a boolean
value and the conversion from string in configuration to boolean did not
happen. Therefore the value from the configuration (eg. 'False') remained
as a string and evaluated to true in boolean context.


fviard:
Issue should not be fixed in MASTER with another fix.
Thank you anyway for your contribution.

PR #803, opened 2016-11-29
Raise warning on filename encoding errors

j0057:
I have a few files with filenames containing non-UTF8 characters, and `s3cmd sync` skips these files. This is what I see with `--debug`:

    INFO: Compiling list of local files...
    DEBUG: Unicodising '03 Radiohead - How Do You\xbf.mp3' using UTF-8
    DEBUG: DeUnicodising u'(1993) Pablo Honey/03 Radiohead - How Do You\ufffd.mp3' using UTF-8

In [S3/FileLists.py:275](https://github.com/s3tools/s3cmd/blob/master/S3/FileLists.py#L275), this file is skipped, because no file with that name can be found on disk. I believe it's useful for s3cmd to not skip the file silently, but to raise a warning, so that the user is aware of the problem and can fix it.

~~If desired, the warning could be raised only if the unicode string contains U+FFFD, the unicode replacement character.~~

### DIFF
--- a/S3/FileLists.py
+++ b/S3/FileLists.py
@@ -276,6 +276,8 @@ def fetch_local_list(args, is_src = False, recursive = None):
                 if not os.path.isfile(deunicodise(full_name)):
                     if os.path.exists(deunicodise(full_name)):
                         warning(u"Skipping over non regular file: %s" % full_name)
+                    elif u'\ufffd' in full_name:
+                        warning(u"Skipping file with encoding problem: %s" % full_name)
                     continue
                 if os.path.islink(deunicodise(full_name)):
                     if not cfg.follow_symlinks:


j0057:
I added the check for U+FFFD, because I'm not sure which other scenarios could lead to skipping a file.

fviard:
In theory it could be good to warn when files are skipped and not only in the case of uffd, but I remember that I noticed a bad side effect when doing that, but I yet can't remember what it was.
In your specific case, is it possible for you to give a try to the following PR that I have so far never merged: https://github.com/s3tools/s3cmd/pull/513

j0057:
Hi Florent,
I retested this scenario using fviard/master@6804801 and also with fviard/replace_more_invalid_chars@ed30953 cherry-picked on top; both revisions still skip the file silently.

I also rebased my branch onto the current master; I still see the correct warnings. I do get another error which also happens on 6804801:

```
  Please try reproducing the error using
  the latest s3cmd code from the git master
  branch found at:
    https://github.com/s3tools/s3cmd
  and have a look at the known issues list:
    https://github.com/s3tools/s3cmd/wiki/Common-known-issues-and-their-solutions
  If the error persists, please report the
  following lines (removing any private
  info as necessary) to:
   s3tools-bugs@lists.sourceforge.net


Invoked as: /home/jjm/git/s3cmd/s3cmd sync --no-progress --stats --cache-file=/some/path/.s3cmd-music-cache /some/path/music/ s3://osme-bucket/Music/
Problem: TypeError: coercing to Unicode: need string or buffer, BadStatusLine found
S3cmd:   1.6.1
python:   2.7.13 (default, Dec 21 2016, 07:16:46)
[GCC 6.2.1 20160830]
environment LANG=en_US.UTF-8

  File "/home/jjm/git/s3cmd/s3cmd", line 2986, in <module>
    rc = main()
  File "/home/jjm/git/s3cmd/s3cmd", line 2905, in main
    rc = cmd_func(args)
  File "/home/jjm/git/s3cmd/s3cmd", line 1831, in cmd_sync
    return cmd_sync_local2remote(args)
  File "/home/jjm/git/s3cmd/s3cmd", line 1805, in cmd_sync_local2remote
    ret = _single_process(args[:-1])
  File "/home/jjm/git/s3cmd/s3cmd", line 1563, in _single_process
    return _child(destination_base, source_args)
  File "/home/jjm/git/s3cmd/s3cmd", line 1667, in _child
    remote_list, dst_exclude_list, remote_total_size = fetch_remote_list(destbase_with_source_list, recursive = True, require_attribs = True)
  File "/home/jjm/git/s3cmd/S3/FileLists.py", line 459, in fetch_remote_list
    objectlist, tmp_total_size = _get_filelist_remote(uri, recursive = True)
  File "/home/jjm/git/s3cmd/S3/FileLists.py", line 396, in _get_filelist_remote
    recursive = recursive, uri_params = uri_params)
  File "/home/jjm/git/s3cmd/S3/S3.py", line 324, in bucket_list
    for truncated, dirs, objects in self.bucket_list_streaming(bucket, prefix, recursive, uri_params, limit):
  File "/home/jjm/git/s3cmd/S3/S3.py", line 355, in bucket_list_streaming
    response = self.bucket_list_noparse(bucket, prefix, recursive, uri_params, max_keys)
  File "/home/jjm/git/s3cmd/S3/S3.py", line 384, in bucket_list_noparse
    response = self.send_request(request)
  File "/home/jjm/git/s3cmd/S3/S3.py", line 1212, in send_request
    warning("Retrying failed request: %s (%s)" % (resource['uri'], unicodise(e)))
  File "/home/jjm/git/s3cmd/S3/Utils.py", line 304, in unicodise
    return unicode(string, encoding, errors)
TypeError: coercing to Unicode: need string or buffer, BadStatusLine found
```

j0057:
This PR has been sitting here for 1.5 years now, ran into another bug, switched to aws-cli now.

PR #805, opened 2016-12-04
setup unittest using moto

shaform:
Previously, I was trying to make s3cmd compatible with Python 3. (#789) However, one thing that made it difficult was that I didn't have constant access to s3 buckets. So I could not test my modifications easily.

This PR attempts to resolve this issue using `moto` as a mock for s3 service. This could also potentially make it possible to add CI services such as [Travis](https://travis-ci.org/) to s3cmd.

Since there are too many existing tests, I only ported a few of the them initially.

### DIFF
--- a/S3/Utils.py
+++ b/S3/Utils.py
@@ -82,10 +82,15 @@ def stripNameSpace(xml):
     """
     removeNameSpace(xml) -- remove top-level AWS namespace
     """
-    r = re.compile('^(<?[^>]+?>\s*)(<\w+) xmlns=[\'"](http://[^\'"]+)[\'"](.*)', re.MULTILINE)
-    if r.match(xml):
-        xmlns = r.match(xml).groups()[2]
-        xml = r.sub("\\1\\2\\4", xml)
+    r = re.compile('^(<?[^>]+?>\s*)?(<\w+) xmlns=[\'"](http://[^\'"]+)[\'"](.*)', re.MULTILINE)
+    matched = r.match(xml)
+    if matched:
+        xmlns = matched.groups()[2]
+        if matched.group(1):
+            xml = r.sub("\\1\\2\\4", xml)
+        else:
+            # moto might not return DTD
+            xml = r.sub("\\2\\4", xml)
     else:
         xmlns = None
     return xml, xmlns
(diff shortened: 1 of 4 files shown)


fviard:
Thank you for your idea.
I didn't knew moto. It could be interesting to test a few things, but I don't find it very great to depend on "moto" and so "boto" for testing. In your PR, you even have to use boto commands.

But for your case, I have a better alternative that I was looking to check in soon for allowing us to test on travis. It is using "minio" compatible s3 server.
When you are on your own machine, you can download it (it's open source but binary are also directly available):
https://www.minio.io/downloads/#minio-server
Then it is really easy to launch it:
chmod +x minio;
./minio server ~/my_folder

Then you have a self contained, mini s3 server that works pretty well (even if there is still a few limitation).
Note: they also provide a free to use online test instance if you want to give it a try: play.minio.io:9000

So, the plan, that they helped me figure out is to download, run minio and then pass the test suite on localhost on travis-ci. I already have a reduced test suite with only non aws specific things tests that i will check in soon.

shaform:
That sounds great. All right, let met close this for now.

fviard:
@shaform I have checked in code for automatic tests with travis.
You can see the status here:
https://travis-ci.org/s3tools/s3cmd
I have activated the testing of python 3.5 even if it is not fatal.
I can add other 3.X versions if that can help you.

Normally you have everything needed to have your own repo testing on his own travis ci.

If you want to performs the tests on you computer, it is almost configuration free for you ;-)

- Download minio:
cd minio; wget -O minio https://dl.minio.io/server/minio/release/linux-amd64/minio; chmod +x minio
- Fix the access and secret key to match the pre made configuration:
export MINIO_SECRET_KEY=zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG
export MINIO_ACCESS_KEY=Q3AM3UQ867SPQQA43P2F
- Prepare you server storage dir:
mkdir server_dir
- Start minio:
./minio server server_dir

Then you can use directly the "-c .travis.s3cfg" config that is checked in master with all your s3cmd.
I recommend to use the "run-test-minio.py" automatic test for tests up to 23 so far.
The following works, but they need a change that i will probably do soon to signature calculation.

ltagliamonte:
Would love to see support for moto, it allows you to run tests also for other aws services, @fviard do you see this happening?

PR #806, opened 2016-12-07
Python3: make relative imports explicit

shaform:
Continue the effort for making it Python 3 compatible.
Some modules (Config, Utils, Progress) have cyclic dependency which makes it difficult to remove their implicit imports.
To keep this PR minimal, I do not attempt to resolve the above issue this time.

### DIFF
--- a/run-tests-minio.py
+++ b/run-tests-minio.py
@@ -7,6 +7,8 @@
 ## License: GPL Version 2
 ## Copyright: TGRMN Software and contributors
 
+from __future__ import print_function
+
 import sys
 import os
 import re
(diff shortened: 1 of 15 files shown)


run-tests-minio.py
```diff
@@ -81,10 +83,10 @@ def is_exe(fpath):
 
 encoding = locale.getpreferredencoding()
 if not encoding:
-    print "Guessing current system encoding failed. Consider setting $LANG variable."
+    print("Guessing current system encoding failed. Consider setting $LANG variable.")
     sys.exit(1)
 else:
-    print "System encoding: " + encoding
+    print(("System encoding: " + encoding))
```
fviard: Is it a typo the 2 parenthesis?

run-tests-minio.py
```diff
@@ -95,7 +97,7 @@ def is_exe(fpath):
     #enc_base_remote = "%s/xyz/%s/" % (pbucket(1), encoding)
     enc_pattern = patterns[encoding]
 else:
-    print encoding + " specific files not found."
+    print((encoding + " specific files not found."))
```
fviard: same a previous chunk, Is it a typo the 2 parenthesis?

fviard:
If you can fix the 2 typos, (i think it is?), I will merge it right after as the rest looks good.

shaform:
Thanks for the review. I've fixed them.

fviard:
Merged. Thanks.
By the way, if you can backport your changes of run-test-minio.py to run-test.py, that would be better for future as run-test.py is the official automatic full test that also cover aws specific things.

shaform:
Surely. I think the only change that's not in ```run-test.py``` is ```cmd_args.insert(0, "python")```. I'll make sure to include that in the next pull request.

PR #808, opened 2016-12-07
Added support for S3 Glacier archive retrieval speed options

robd003:
Users can now specify whether they want their S3 Glacier retrieval to be costly and quick or slow and cheap.


fviard:
Merged. Good contribution. Thank you very much and thank you also for the fixes in help strings.

PR #809, opened 2016-12-09
Python3: fix several imports, and run 2to3 on s3cmd

shaform:
Continue the effort for making s3cmd Python 3 compatible.

- Remove remaining implicit relative imports
I thought about removing the circular dependencies, but in the end I decided to simply modify the imports to minimize changes.
- Fix some imports that have different names in Python 3.
- Run 2to3 on s3cmd to fix several compatibility issues

### DIFF
--- a/S3/Utils.py
+++ b/S3/Utils.py
@@ -6,6 +6,8 @@
 ## License: GPL Version 2
 ## Copyright: TGRMN Software and contributors
 
+from __future__ import absolute_import
+
 import os
 import sys
 import time
@@ -15,7 +17,7 @@ import random
 import errno
 from calendar import timegm
 from logging import debug, warning, error
-from ExitCodes import EX_OSFILE
+from .ExitCodes import EX_OSFILE
 try:
     import dateutil.parser
 except ImportError:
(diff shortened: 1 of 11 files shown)


s3cmd
```diff
@@ -2110,9 +2116,9 @@ def cmd_fixbucket(args):
                 debug("Step 3:  ... then to %s" % key_new)
                 src = S3Uri("s3://%s/%s" % (culprit.bucket(), key_bin))
                 dst = S3Uri("s3://%s/%s" % (culprit.bucket(), key_new))
-		if cfg.dry_run:
-		    output("[--dry-run] File %r would be renamed to %s" % (key_bin, key_new))
-		    continue
```
shaform: Notice that here are some strange TABs.

fviard:
Thanks.
Before merging, there is just a point that annoys me, but I don't know if it results of the usage of 2to3 or if it is needed by python 3 for a reason. Please see my questions inline.

S3/Utils.py
```diff
@@ -33,8 +35,8 @@
     sys.stderr.flush()
     sys.exit(EX_OSFILE)
 
-import Config
-import Exceptions
+import S3.Config
```
shaform: These are needed because of circular import dependency. If not for this reason `from . import Config` is certainly better.

shaform: In the long run, it might be better to remove all those circular import dependencies. But doing so might change some of the APIs.

fviard:
Thank you!

PR #828, opened 2017-01-22
filter out non-clacier or clacier which already restored

wangzhihao:
Filter out non-GLACIER  files and GLACIER files which already start or finish restored.

### DIFF
--- a/s3cmd
+++ b/s3cmd
@@ -765,13 +765,16 @@ def cmd_object_restore(args):
         uri = S3Uri(item['object_uri_str'])
         if not item['object_uri_str'].endswith("/"):
             try:
-                response = s3.object_restore(S3Uri(item['object_uri_str']))
-                output(u"restore: '%s'" % item['object_uri_str'])
-            except S3Error as e:
-                if e.code == "RestoreAlreadyInProgress":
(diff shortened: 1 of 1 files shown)


fviard:
The idea is interesting, but isn't it more efficient to simply still send only the "restore" request and ignore based on the error/reply? Maybe the "output" message could be fixed to only say "restore" for files that will in effect be restored.
In the python way "ask forgiveness, not permission".
As in your PR, you suggest to issue a "get_info" first for each file.

wangzhihao:
Yeah, error-and-reply sounds more graceful. But my main concern is that the error code `InvalidObjectState` is not clear to indicate non-glacier. Is it safe for us to ignore all `InvalidObjectState` error?
```python
   if e.code == "InvalidObjectState":
                 warning("%s: %s" % (e.message, item['object_uri_str']))
```

fviard:
Sorry that I forgot to reply since so long, i don't know if you are still looking at this MR, but the reply is yes, skipping all InvalidObjectState looks ok to me.
We can probably have a more explicit message prefix in the warning also ("Skipping as not a Glacier object or not in a restorable state:")

PR #835, opened 2017-02-15
Use http.client for Python 3

alexjh:
The custom httplib isn't needed for Python 3 as it supports Continue.

https://github.com/python/cpython/blob/f15fa87e5a66c0000f63be5065ddee4a52d5660f/Lib/http/client.py#L294

This is untested as there's a lot more Python 3 changes to make, but I'm hoping to make some progress on that.

### DIFF
--- a/S3/ConnMan.py
+++ b/S3/ConnMan.py
@@ -9,7 +9,11 @@
 from __future__ import absolute_import
 
 import sys
-from .custom_httplib27 import httplib
+try:
+    # python 3 support
+    import http.client as httplib
+except ImportError:
+    from .custom_httplib27 import httplib
 import ssl
 from threading import Semaphore
 from logging import debug


fviard:
Thank you for your contribution, but httplib still doesn't support the real "continue" feature in python3.
The behavior in py3 is the same as in py2:
a "continue" status code is ignored but for example it is not propagated up. That prevents for example to detect "redirections" and things like that.

But don't worry, i did the py3 version of the custom_httplib and a few other changes to be more python3 compatible that I will be soon be pushed.
Py3 support is more complicated than it can looks because a lot of small behavioral things have changed.

alexjh:
Ok, thanks for getting to this so quickly.

I'd be happy to help test any Py3 work in the future.

fviard:
@alexjh I did push in master my fix for the custom httplib and a lot of others.
Don't hesitate to start to play with that ^^

PR #837, opened 2017-02-22
Ensure testsuite-out is created before running dd

alexjh:
If you're running single tests, they will fail on test 15 because the testsuite-out dir isn't created.


fviard:
Merged, thank you.
For a future new PR, please of also updating run-test-minio.py.

PR #851, opened 2017-03-31
Fixes #849 by adding requester pays header to HEAD requests

### LINKED ISSUE #849: Requester-pays multipart downloads fail because of HEAD requests
#### Context
In s3cmd, [multipart file downloads](https://github.com/s3tools/s3cmd/blob/master/S3/FileLists.py#L428) require [HEAD requests](https://github.com/s3tools/s3cmd/blob/master/S3/S3.py#L927) that fail because they [do not include the requester-pays flag](https://github.com/s3tools/s3cmd/blob/master/S3/S3.py#L144).

This requirement is explained in the [AWS docs for RequesterPays buckets](http://docs.aws.amazon.com/AmazonS3/latest/dev/RequesterPaysBuckets.html):

> After you configure a bucket to be a Requester Pays bucket, requesters must include x-amz-request-payer in their requests either in the header, for **POST, GET and HEAD** requests, or as a parameter in a REST request to show that they understand that they will be charged for the request and the data download.

#### Solution
Add the requester-pays header to HEAD requests.

MrChrisRodriguez:
See: https://github.com/s3tools/s3cmd/issues/849

“After you configure a bucket to be a Requester Pays bucket, requesters must include x-amz-request-payer in their requests either in the header, for POST, GET and **HEAD** requests, or as a parameter in a REST request to show that they understand that they will be charged for the request and the data download.”

http://docs.aws.amazon.com/AmazonS3/latest/dev/RequesterPaysBuckets.html

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -142,7 +142,7 @@ class S3Request(object):
         self.requester_pays()
 
     def requester_pays(self):
-        if self.s3.config.requester_pays and self.method_string in ("GET", "POST", "PUT"):
+        if self.s3.config.requester_pays and self.method_string in ("GET", "POST", "PUT", "HEAD"):
             self.headers['x-amz-request-payer'] = 'requester'
 
     def update_timestamp(self):


fviard:
That makes sense, merged, thank you very much!

PR #855, opened 2017-04-06
fix empty directory check on Windows

nathantreid:
Directories on Windows use `\` instead of `/`. This causes the following error, which is fixed by this PR:
```
ERROR: Download of 'mydir/' failed (Reason: Unknown OsError 17)
ERROR: Exiting now because of fatal error
ERROR: [Error 183] Cannot create a file when that file already exists
```


fviard:
Thank you for the PR with the root cause analysis, some changes were pushed in MASTER to better handle directories with s3 and this should also resolve this issue.

PR #856, opened 2017-04-07
run-tests.py: use lowercase bucket name

guygurari:
S3 does not support uppercase letters in bucket names. The bucket name
used for testing includes the user name, so convert it to lower case first.


guygurari:
I included this small change in another pull request, so no need to add it twice.

Issue #854, opened 2017-04-05
Data loss when moving multiple files


guygurari:
I would like to move all tar files (`1.tar`, `2.tar`, ...) into a subdirectory `dir`, so I run the following.

`s3cmd mv s3://bucket/*.tar s3://bucket/dir`

In practice this takes each tar file and renames it to a file called `dir`. This leads to data loss because each rename operation overwrites the file created by the previous rename.

Data loss occurs because S3 allows both a directory called `dir` and a file called `dir` to exist at the same time. The command above expands into the following sequence of commands.

`s3cmd mv s3://bucket/1.tar s3://bucket/dir`
`s3cmd mv s3://bucket/2.tar s3://bucket/dir`
`s3cmd mv s3://bucket/3.tar s3://bucket/dir`
`...`

Each tar file gets renamed `dir`, and each subsequent command overrides the file created by the previous one.

A workaround is to run the command with a trailing slash, like this.

`s3cmd mv s3://bucket/*.tar s3://bucket/dir/`

guygurari:
The problem seems to be in `s3cmd` in the function `subcmd_cp_mv`:

```
for key in remote_list:
    if destination_base.endswith("/"):
        remote_list[key]['dest_name'] = destination_base + key
    else:
        remote_list[key]['dest_name'] = destination_base
```

If there is no trailing slash, the files all get moved to `destination_base`. To me, this seems like unexpected and dangerous behavior. For comparison, UNIX `mv` would give an error in a similar situation (trying to rename multiple files into a single file). I'd be happy to add a pull request that fixes this, but is there a good reason for the existing behavior?

fviard:
Thank you for this report.
indeed that behavior is not very great.

Something should probably done like in "cmd_sync_remote2local" where an error is returned if you try to use multiple source with a destination that has no "/".

As you noticed, a big issue is that we don't know if source/dest are directories or files.
Looking at the code in subcmd_cp_mv, that is currently a little more complicated than this short snippet of code. If the "recursive" flag is set by config or command line, there is one behavior, otherwise there is another behavior that is the one that you noticed.

It is not great if you want my opinion, but before changing that we have to balance "the future" and the "legacy" of people scripts/usage that could rely on the current behavior.

Patches are very welcome.
Just try to evaluate all the cases that are possibles: multiple sources that are dirs; that are files; just one files that is "renamed" etc...

guygurari:
[...]

PR #857, opened 2017-04-08
address issue #854 : don't allow mv/cp of multiple files to single file

guygurari:
When trying to cp/mv multiple files, make sure the destination
has a trailing slash so it is treated as a directory. Otherwise,
multiple files can get moved to a single file, leading to data
loss.

### DIFF
--- a/run-tests-minio.py
+++ b/run-tests-minio.py
@@ -397,6 +397,32 @@ f.close()
 ## ====== Clean up local destination dir
 test_flushdir("Clean testsuite-out/", "testsuite-out")
 
+## ====== Moving things without trailing '/'
+os.system('dd if=/dev/urandom of=testsuite-out/urandom1.bin bs=1k count=1 > /dev/null 2>&1')
+os.system('dd if=/dev/urandom of=testsuite-out/urandom2.bin bs=1k count=1 > /dev/null 2>&1')
+test_s3cmd("Put multiple files", ['put', 'testsuite-out/urandom1.bin', 'testsuite-out/urandom2.bin', '%s/' % pbucket(1)],
+           must_find = ["%s/urandom1.bin" % pbucket(1), "%s/urandom2.bin" % pbucket(1)])
+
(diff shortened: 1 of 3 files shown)


fviard:
Regarding the fix for the tests, that is good, but that would be great if you can also backport your changes to run-tests-minio.py. For the moment, it is a little duplicates, but it is the "minio" one that is run as "automatic" tests by travisci.

Regarding the fix itself, i did think more about this and in my opinion, the best would be to throw something like: (remote2local)
raise ParameterError("Destination must be a directory and end with '/' when downloading multiple sources.")
When you encounter the bad case. So things are explicit and not implicit.
And with that, maybe you can get ride of the "cfg.recursive" and have a single behavior for all the case.
in1 out -> ok
in1 out/ -> ok
in1/ out -> wrong, raise error
in1 in2 out -> wrong, raise error
in1 in2 out/ -> ok
in1/ out/ -> ok

People currently using multiple sources with recursive and no "/" for destination could encounter a behavior change. But it would be an explicit error and the fix for them would be to add a "/" at destination, so I think it should be ok.

guygurari:
Thanks for looking at this. I changed the behavior to raise an exception when trying to move 'in1 in2 -> out' or when moving recursively to 'out' (without a '/'). I also added the tests to `run-tests-minio.py`.

I did not see a way to ignore `cfg.recursive`, because moving `in -> out` should be allowed when it moving non-recursively. Also, the case 'in1/ -> out' is already handled in `S3.FileLists` before it reaches this code, so I did not handle it.

fviard:
I'm sorry to take a little time to reply to you, i have a little back log that accumulated.
Thank you for the update.
I will review that in the coming days.

guygurari:
No worries, I know we're all busy :-)

fviard:
Merged! Thank you very much for your contribution, to have taken care of adding auto tests and for your patience!

guygurari:
Great! Glad I could contribute.

PR #860, opened 2017-04-12
Remove guess on region

### LINKED ISSUE #859: s3cmd is hard to use with custom s3 implementation
Right now, `s3cmd` implement some AWS S3 conventions that are not required and make other implementations failing.

For example, impossible to use a bucket named `eu` (it's upper cased to `EU`), or a bucket named `us`, `LocationConstraint` is just ignored.

guilhem:
This prevent s3cmd to fail with other s3 implementations

Fix #859

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -393,12 +393,8 @@ class S3(object):
     def bucket_create(self, bucket, bucket_location = None):
         headers = SortedDict(ignore_case = True)
         body = ""
-        if bucket_location and bucket_location.strip().upper() != "US" and bucket_location.strip().lower() != "us-east-1":
+        if bucket_location:
             bucket_location = bucket_location.strip()
-            if bucket_location.upper() == "EU":
-                bucket_location = bucket_location.upper()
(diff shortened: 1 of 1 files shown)


benhowell:
:+1: 
Agree. I patched s3cmd myself to make it work with DigitalOcean spaces and came here to make a pull request. But here it is already ;-)

fviard:
@guilhem @benhowell Sorry, I didn't understood this issue and the proposed PR, but I hadn't time to look deeper into this issue before.

Based on @benhowell, I opened a DO space to try s3cmd with it to try to reproduce any issue, but I can't. So I supposed that it was patched on their side.

Is it possible for one of you to describe me in details that was the issue/root cause, and why this PR would have fixed that? (even if maybe it is fixed with DO).
I look at #859 but I don't understand how the "location" value as any relation with naming a bucket "eu" or "us"? My best guess would be that you use to have a dynamic "location" field in your host_bucket config where "regions" are not supported by DO.

PR #917, opened 2017-09-06
Adding config option for signurl to use https prefix

thelan:
Hello,

This is just a quick fix to add an configuration option when using sign url to generate an HTTPS link


fviard:
Hi,
Thanks for this PR. That looks good to me, but there is just a little thing that i would like to be changed.
For being consistent with the "use_https" option, and the cmdline arg "signurl", I think that it would be better to have the following name for this option:
signurl_use_https

thelan:
i've fixed the option name

fviard:
Merged, thank you!

Issue #880, opened 2017-06-13
ETag shouldn't be used for MD5 verification


EgoAleSum:
This is connected with this issue I've filed for Minio: https://github.com/minio/minio/issues/4537

s3cmd considers the ETag field to be the file's md5, but  "the ETag may or may not be an MD5 digest of the object data" (see: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html ) This is causing issues with tools such as Minio in gateway mode: `s3cmd put` always fails with checksum errors.

The `--no-check-md5` option doesn't seem to work with the `put` command.

krishnasrinivas:
s3cmd devs, for MD5 verification can you guys use `Content-MD5` header instead of ETag? When `minio` is run in s3-gateway mode it makes Azure Blob Storage look like S3. Azure stores ETag that looks like `0x8D4B2AE97754E88` which is not MD5.

According to RFC https://tools.ietf.org/html/rfc7232#section-2.3 ETag is an opaque string. AWS-S3 for single part PUT just uses MD5 for ETag. For multipart it's not MD5.

Using Content-MD5 header instead of ETag for MD5 verification can make s3cmd work with Azure Blob Storage using minio.

fviard:
Hum, interesting question.
@EgoAleSum the --no-check-md5 doesn't work in your case because it is designed to avoid the cost of calculating md5 of all the files to select what to update in a sync, but not the less expensive one of ensuring that files were correctly transferred

@krishnasrinivas The idea of using content-md5 is interesting, but i'm concerned with the following point:
- Would most of the s3 custom servers implementations support this? How to know if it is not the case and so still spot corruption? There could probably be an option flag, but that is really implementation dependent and would not means a lot to normal users. 

Currently, too much in the S3 protocol assume that the etag is the hash except some special cases:
http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html

At this subject, I'm curious to know how you dealt with "multipart uploads" as you are supposed to complete them with the list of etags?

@krishnasrinivas If modifications are possibles on your side, I would suggest an alternative:
If you don't have access to the md5 from azure and so that you can't "replace" the etags with it on the fly, what do you think about doing the following:
- Append a "-1" to the end of the etag coming from azure. Like (0xDE32425BFA-1).
(Maybe even remove the 0x at the beginning).
[...]

PR #923, opened 2017-10-08
Don't need to compute ETag md5sum match if it is not a real md5sum

harshavardhana:
Fixes https://github.com/s3tools/s3cmd/issues/880


fviard:
Merged, thank you!

PR #925, opened 2017-10-10
Change the marker value for NextMarker in bucket_list_streaming

### LINKED ISSUE #924: Use NextMarker unless object name when listing objects
S3 protocol support `NextMarker` and an object name in the marker parameter, but Azure expects marker (https://docs.microsoft.com/en-us/rest/api/storageservices/List-Blobs?redirectedfrom=MSDN) in the ListBlobs request to be the NextMarker field of the previous ListBlobs result.

I purpose to use the NextParameter in `bucket_list_streaming` for adding the compatibility with Minio.

For more information, you can check this minio issue: https://github.com/minio/minio/issues/4948

jmaitrehenry:
Fix #924

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -328,6 +328,9 @@ class S3(object):
         def _get_common_prefixes(data):
             return getListFromXml(data, "CommonPrefixes")
 
+        def _get_next_marker(data, current_list):
+            return getTextFromXml(response["data"], "NextMarker") or current_list[-1]["Key"]
+
         uri_params = uri_params and uri_params.copy() or {}
         truncated = True
         prefixes = []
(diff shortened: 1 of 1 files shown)


fviard:
Thanks for the PR.
The idea of it looks ok to me but a fast internet search let me think that there is still a risk that the "nextmarker" would be absent or empty.

So, I think that it should probably be better to do something like:
maker = "get nextMarker", if not or empty, marker = current key[-1] (get last key as before).

(just cosmetic, but for consistency it could also be better to add a "_get_next_marker" like the "_get_contents".)

jmaitrehenry:
@fviard done

fviard:
Merged, thank you!

PR #926, opened 2017-10-19
Return early when modify metadata with not correct key

gaosibei:
Signed-off-by: Sibei Gao <gaosb@inspur.com>


fviard:
Hi,
I can be wrong but the modification that you propose is not right.
There are usages for a "modify" that will not change a "custom" header.
For example if you want to change the content-type or any vendor custom header.

What was the intended purpose of your PR? you wanted to "typos" like forgetting the x-amz-meta?

PR #938, opened 2017-12-01
Fixes bug in cinvalinfo when only one path is supplied to cloudfront invalidation request.

Nossnevs:
Fixes bug in cinvalinfo when only one path is supplied. Before it displayed the nr of characters in the string instead of "nr of paths: 1".


fviard:
Nice catch!
Merged, thank you.

Issue #525, opened 2015-04-09
--cf-invalidate doesn't invalidate copied files


rstacruz:
Let's say you have 3 identical files:

```
/files/v0/file.js
/files/v0.1/file.js
/files/v0.1.9/file.js
```

and you sync it using `s3cmd sync`:

```
s3cmd sync --cf-invalidate --acl-public --preserve --recursive ./files s3://my-cdn/

WARNING: Module python-magic is not available. Guessing MIME types based on file extensions.
./files/v0.1.9/file.js -> s3://my-cdn/files/v0.1.9/file.js  [1 of 1]
 37842 of 37842   100% in    1s    27.05 kB/s  done
remote copy: files/v0.1.9/file.js -> files/v0.1/file.js
remote copy: files/v0.1.9/file.js -> files/v0/file.js
Done. Uploaded 37842 bytes in 2.9 seconds, 12.81 kB/s. Copied 2 files saving 75684 bytes transfer.
Created invalidation request for 1 paths
Check progress with: s3cmd cfinvalinfo cf://.../...
```

It says it `Created invalidation request for 1 paths`, which should probably be 3.

rstacruz:
anyone got a workaround or resolution for this?

fviard:
Should be fixed with:
https://github.com/s3tools/s3cmd/commit/aa18fa41cd4e63a015ab75bc0ccf99d8d49be151

PR #939, opened 2017-12-07
Fixed a bug which didn't invalidate any files which were copied remotely during sync

kgritesh:
Fixed a bug which didn't invalidate any files which were copied remotely during sync. Fixes issue #525


fviard:
Following your proposal, I did an improved version that I think would be a little more efficient:
https://github.com/s3tools/s3cmd/commit/aa18fa41cd4e63a015ab75bc0ccf99d8d49be151
But thank you very much for having narrowed down this issue and the PR!

PR #940, opened 2017-12-07
Fix handling of config tri-state bool values (like acl_public)

bcl:
Commit bc38c2a879305ffd3cdd4f025f4161b430e79150 added the ability to set
acl_public from the config file -- but it did not take into account that
it is a 'tri-state' value so the string is what is written, causing it
to ALWAYS be true.

This results in unexpected and insecure behavior if 'acl_public = False'
is used in the config file.

This patch modifies the Config.update_option to check the option type
AND the value if the original is None. If an option's default is None
and the value is a bool (true, false, yes, no, 0, 1) then it will set it
to a bool value.

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -44,6 +44,28 @@ def config_unicodise(string, encoding = "utf-8", errors = "replace"):
     except UnicodeDecodeError:
         raise UnicodeDecodeError("Conversion to unicode failed: %r" % string)
 
+def is_true(value):
+    """Check to see if a string is true, yes, on, or 1
+
+    value may be a str, or unicode.
+
+    Return True if it is
+    """
+    return value.lower() in ("true", "yes", "on", "1")
+
+def is_false(value):
+    """Check to see if a string is false, no, off, or 0
+
+    value may be a str, or unicode.
+
+    Return True if it is
+    """
+    return value.lower() in ("false", "no", "off", "0")
+
+def is_bool(value):
+    """Check a string value to see if it is bool"""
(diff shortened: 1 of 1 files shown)


bcl:
This should also solve #793 but in a more generic way.

fviard:
Thank you for having catch that.
And the proposed fix is quite graceful!

But I still see a little drawback, if you can add a fix for that in your PR, that would be awesome:
Now, all the fields having a "None" as default config will be "treated" by your new code.
So 2 sides effects:
- One that is in fact a text field, might have boolean set for it if by bad luck or else "true" or "1" be used for example. (A rare corner case i guess)
- Any provided text for such an option will go trough your "is_bool" function. As it is using a "str(...)" construct, chances are that this will crash with utf8 values and python2.7. (That case is likely for someone using content_disposition).

Possible fix:
I see 3 options that are like that in the config: content_disposition, content_type, and upload_id.
A fast lookup of the code let me think that they may have their default value replaced to "".
A few lines of code will have to be updated to not look "is None" but just if "None or empty".
Luckily, i don't see any case where these 3 fields could have a legitimate "" value that has to be used.

bcl:
I've updated the PR with a small change, dropping use of str() and renaming the functions. From what I can see they will always be called with unicode so it can just use .lower() on that. And if it ever gets called with a str, .lower() still works.

I don't think the corner case of using a bool string for one of the other entries is going to be a problem, but if you want to deal with that separately I'll leave that up to you.

fviard:
Finally merged, thank you. Sorry for the long delay.

PR #947, opened 2018-01-13
Fix Python 3 bytes string encoding when getting IAM credentials

alexallah:
The `files` variable has a type of `bytes()` in Python 3.

For example, if `files = b'master'` then the request path would be formatted as
`"/latest/meta-data/iam/security-credentials/b'master'"`

The `.decode()` method converts a bytes into a proper string in Python 3 and doesn't change anything in Python 2.

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -217,7 +217,7 @@ class Config(object):
             resp = conn.getresponse()
             files = resp.read()
             if resp.status == 200 and len(files)>1:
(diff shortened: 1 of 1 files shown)


alexallah:
Without the fix, I was getting this error message:
```
ERROR: /root/.s3cfg: None
ERROR: Configuration file not available.
ERROR: Consider using --configure parameter to create one.
```

fviard:
Merged, thank you!

Issue #906, opened 2017-07-31
$AWS_PROFILE is not honored


matt-pilgrim:
[This](https://stackoverflow.com/questions/33551537/how-to-set-the-profile-when-running-the-s3cmd-sync-command) mentions something similar, and suggests using `--configure`, but that apparently uses the `~/.s3cfg` file.  Shouldn't I be able to use a command like `AWS_PROFILE=foo s3cmd ...` and use the "foo" profile from `~/.aws/credentials`?  

As someone who routinely uses multiple aws profiles, this is painful because I don't really want to keep another file in sync, and I also don't want to have to view my credentials files and copy/paste environment variables for AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY.

 If there is agreement that this is a bug, I can probably submit a patch.

EDIT:  found [this](https://github.com/s3tools/s3cmd/issues/104#issuecomment-90974616) comment from 2015 suggesting using the credentials file, and [this](https://github.com/s3tools/s3cmd/issues/104#issuecomment-93788313) mention in the same thread that a patch would be accepted.  Can anyone confirm this is still desired?

felipe1982:
If like this feature. I'm using `awsume` in the meantime which creates environment variables 

_Sent from my Google Pixel using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

alexanderkjeldaas:
@felipe1982 which awesome command do you use?  awesome doesn't set any environment variables for me.

felipe1982:
That's only what awsume does: sets environment variables. Make sure you have the Alias set, or run `. awsume `

fviard:
Should be already fixed with merged PR #966 .

Issue #570, opened 2015-06-16
Support ~/.aws/credentials


jjshoe:
It would be great if s3cmd would check ~/.aws/credentials (http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html) for just the credentials to use.

mdomsch:
Export AWS_CREDENTIAL_FILE=~/.aws/credentials, and then run s3cmd.  This is
in upstream master branch and v1.5.2.

commit dc590d62a12f309802a9865d34204117a8defebe
Author: David Kohen <kohen.d@gmail.com>
Date:   Wed Feb 13 15:57:11 2013 +0200

    Add support for IAM roles and temp tokens

    Including:
    Refresh temp tokens on every request.
    Add support for roles in S3, CloudFront and SimpleDB.
    Add support for AWS_CREDENTIALS_FILE env variable and the file it
references

jjshoe:
@mdomsch very handy, though not the same thing by any means. The point was to check the location that AWS recommends for you to store your credentials.

mdomsch:
I'm not sure why the location was defined in an environment variable
instead of simply looked at.  David?

jjshoe:
http://blogs.aws.amazon.com/security/post/Tx3D6U6WSFGOK2H/A-New-and-Standardized-Way-to-Manage-Credentials-in-the-AWS-SDKs

PR #966, opened 2018-04-28
Issue #906 add support for $AWS_PROFILE

Trane9991:
Add support for reading credentials from default, standardized AWS credentials file [1].

New logic works as in PR #570 but with additional changes.

Reading credentials from the AWS credentials file can be triggered by setting any of following env variables:
- `AWS_CREDENTIAL_FILE`
- `AWS_PROFILE`

Default value for `AWS_PROFILE` is `default` and for `AWS_CREDENTIAL_FILE` is `~/.aws/credentials`. 

If `AWS_CREDENTIAL_FILE` set to non-existing file, it will fallback to the default value. This way you can use `AWS_CREDENTIAL_FILE=true` if you want to read credentials from the default location.

[1] https://aws.amazon.com/blogs/security/a-new-and-standardized-way-to-manage-credentials-in-the-aws-sdks/

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -23,6 +23,7 @@ try:
 except ImportError:
     import http.client as httplib
 import locale
+import configparser
 
 try:
     unicode
@@ -211,8 +212,8 @@ class Config(object):
             try:
                 self.read_config_file(configfile)
             except IOError:
-                if 'AWS_CREDENTIAL_FILE' in os.environ:
-                    self.env_config()
+                if 'AWS_CREDENTIAL_FILE' in os.environ or 'AWS_PROFILE' in os.environ:
+                    self.aws_credential_file()
 
             # override these if passed on the command-line
             if access_key and secret_key:
@@ -275,38 +276,35 @@ class Config(object):
             except:
                 warning("Could not refresh role")
 
-    def env_config(self):
-        cred_content = ""
+    def aws_credential_file(self):
         try:
-            cred_file = open(os.environ['AWS_CREDENTIAL_FILE'],'r')
-            cred_content = cred_file.read()
-        except IOError as e:
-            debug("Error %d accessing credentials file %s" % (e.errno,os.environ['AWS_CREDENTIAL_FILE']))
-        r_data = re.compile("^\s*(?P<orig_key>\w+)\s*=\s*(?P<value>.*)")
-        r_quotes = re.compile("^\"(.*)\"\s*$")
-        if len(cred_content)>0:
-            for line in cred_content.splitlines():
-                is_data = r_data.match(line)
-                if is_data:
-                    data = is_data.groupdict()
-                    if r_quotes.match(data["value"]):
-                        data["value"] = data["value"][1:-1]
-                    if data["orig_key"] == "AWSAccessKeyId" \
-                       or data["orig_key"] == "aws_access_key_id":
-                        data["key"] = "access_key"
-                    elif data["orig_key"]=="AWSSecretKey" \
-                       or data["orig_key"]=="aws_secret_access_key":
-                        data["key"] = "secret_key"
-                    else:
-                        debug("env_config: key = %r will be ignored", data["orig_key"])
+            config = configparser.ConfigParser()
 
-                    if "key" in data:
-                        Config().update_option(data["key"], data["value"])
-                        if data["key"] in ("access_key", "secret_key", "gpg_passphrase"):
-                            print_value = ("%s...%d_chars...%s") % (data["value"][:2], len(data["value"]) - 3, data["value"][-1:])
-                        else:
-                            print_value = data["value"]
-                        debug("env_Config: %s->%s" % (data["key"], print_value))
+            aws_credential_file = os.path.expanduser('~/.aws/credentials') 
+            if 'AWS_CREDENTIAL_FILE' in os.environ and os.path.isfile(os.environ['AWS_CREDENTIAL_FILE']):
+                aws_credential_file = os.environ['AWS_CREDENTIAL_FILE']
+
+            debug("Reading AWS credentials from", aws_credential_file)
+            config.read(aws_credential_file)
+            profile = "default"
+            if 'AWS_PROFILE' in os.environ:
+                profile = os.environ['AWS_PROFILE'] 
+
+            profile_access_key = config.get(profile, 'aws_access_key_id')
+            profile_secret_key = config.get(profile, 'aws_secret_access_key')
+            self.access_key = config_unicodise(profile_access_key)
+            self.secret_key = config_unicodise(profile_secret_key)
+
+            try:
+                profile_access_token = config.get(profile, 'aws_session_token')
+                Config().access_token = config_unicodise(profile_access_token) 
+            except configparser.NoOptionError:
+                pass
+
+        except IOError as e:
+            error("%d accessing credentials file %s" % (e.errno,os.environ['AWS_CREDENTIAL_FILE']))
+        except (configparser.NoOptionError, configparser.NoSectionError) as e:
+            error(e)
 
     def option_list(self):
         retval = []


Trane9991:
Hello @fviard,
Could you take a look at this PR, when possible?

fviard:
@Trane9991 Thank you for your contribution, and sorry for my slow response time.
I wanted to take the time to read carefully to ensure that it doesn't break anything.
But the direction of the patch is good.

A few things I see:
-
            error("%d accessing credentials file %s" % (e.errno,os.environ['AWS_CREDENTIAL_FILE']))
It would probably be better to keep it as it was before.
The thing can be open for discussion, but I think that if no aws configure can be loaded, that is not really an error. That could be a normal case.
Maybe that can be just a warning? more explicit, something like no aws credential file found or found but can't be loaded.

- 
            error(e)
at the end. Don't do this, always put an explicit error message, because otherwise some orphan errors would appear in the log and no one will understand where it came from.
As for the previous one, as we just pass without caring, maybe it should be only an warning or debug.

-
            It would probably be better to keep the log that was done of "keys" that are overriden. That allows to understand where the keys that are in use come from.

- 
           It would be better to only set config variable through the update_config like it was done before. Some treatment can be done on the config by it.

- 
           if data["orig_key"] == "AWSAccessKeyId" \	
                       or data["orig_key"] == "aws_access_key_id":
I don't know a lot of history of s3cmd, but I think that if there was this line, it is probably that there is another aws config file or a previous version of its format that used such a key?
Would be good to understand where it came from to be sure to not break support for something existing.

- 
This one is just a matter of taste I guess, but for the import i would have preferred something like : 
try: 
 from configparser import NoOption, NoSection, ConfiParser as PyConfigParser
except importerror:
  # Python2 fallback code
  from Configparser import ....

And so you can use them directly in the code.
the name PyConfigParser is to be very very explicit to not think by accident that defaultconfigparser is a kind of object of s3cmd related to s3cmd configparser


I hope that you will still like to complete your PR despite this long review ^^

Trane9991:
Thanks for the in-depth review, the advice is highly appreciated!

Regarding the:
*     error("%d accessing credentials file %s" % (e.errno,os.environ['AWS_CREDENTIAL_FILE']))
The idea was if person specified `AWS_CREDENTIAL_FILE` then he/she wanted to read credentials from the `AWS_CREDENTIAL_FILE`, and if it wasn't found, then it is an error. But if you still think it should be `warning` it is fine with me as well :)

*     if data["orig_key"] == "AWSAccessKeyId" \	
                 or data["orig_key"] == "aws_access_key_id":
I'm not sure about this one as well, but it was added a long time ago since then credentials file format was standardized and usage of the credentials in the five years old and the unsupported format should be considered as a bad practice, and not sure if it needs to be supported.

S3/Config.py
```diff
@@ -275,38 +281,46 @@ def role_refresh(self):
...
 
-    def env_config(self):
-        cred_content = ""
+    def aws_credential_file(self):
         try:
-            cred_file = open(os.environ['AWS_CREDENTIAL_FILE'],'r')
-            cred_content = cred_file.read()
+            aws_credential_file = os.path.expanduser('~/.aws/credentials') 
+            if 'AWS_CREDENTIAL_FILE' in os.environ and os.path.isfile(os.environ['AWS_CREDENTIAL_FILE']):
+                aws_credential_file = os.environ['AWS_CREDENTIAL_FILE']
```
fviard: I know that it was not done before, but as an improvement, it would be great if you can ensure to pass the strings you get from os.environ to config_unicodise.
The tests to be done to check if things are good, is like to create a folder with chinese characters, and try to put some chinese chars inside the configuration file itself. Normally keys don't have special chars, but so we are sure that the code is proof to nasty issues in the future.

S3/Config.py
```diff
@@ -275,38 +281,46 @@ def role_refresh(self):
...
+
+            config = PyConfigParser()
+
+            debug("Reading AWS credentials from", aws_credential_file)
+            config.read(aws_credential_file)
+
+            profile = "default"
+            if 'AWS_PROFILE' in os.environ:
+                profile = os.environ['AWS_PROFILE']
```
fviard: 2 previous lines can be simplified in to:
profile = os.environ.get('AWS_PROFILE', 'default')
So with the unicode safe conversion:
profile = config_unicodise(os.environ.get('AWS_PROFILE', 'default')

S3/Config.py
```diff
@@ -275,38 +281,46 @@ def role_refresh(self):
...
+            # trying to read aws_access_key_id from credentials file
+            profile_access_key = config.get(profile, 'aws_access_key_id')
+            debug('Setting "aws_access_key_id" from file {} as "access_key"'.format(aws_credential_file))
+            Config().update_option('access_key', config_unicodise(profile_access_key))
+
+            # trying to read aws_secret_access_key from credentials file
+            profile_secret_key = config.get(profile, 'aws_secret_access_key')
+            debug('Setting "aws_secret_access_key" from file {} as "secret_key"'.format(aws_credential_file))
+            Config().update_option('secret_key', config_unicodise(profile_secret_key))
+
```
fviard: Based on the previous code behavior, I think that you should take care that maybe only a part of access/secret/token keys could be stored in the config file.
Like only the session token, or only the access_key, ...
The rest being provided on the command line for example.

S3/Config.py
```diff
@@ -275,38 +281,46 @@ def role_refresh(self):
...
-                    if "key" in data:
-                        Config().update_option(data["key"], data["value"])
-                        if data["key"] in ("access_key", "secret_key", "gpg_passphrase"):
-                            print_value = ("%s...%d_chars...%s") % (data["value"][:2], len(data["value"]) - 3, data["value"][-1:])
-                        else:
-                            print_value = data["value"]
-                        debug("env_Config: %s->%s" % (data["key"], print_value))
+            error("%d accessing credentials file %s" % (e.errno, aws_credential_file))
+        except NoOptionError as e:
+            error("Couldn't find required key '{}' for the AWS Profile '{}' in the credentials file '{}'".format(e.option, e.section, aws_credential_file))
```
fviard: As discussed, "warning" would probably be enough. For the next exception also.

S3/Config.py
```diff
@@ -275,38 +281,46 @@ def role_refresh(self):
...
-            for line in cred_content.splitlines():
-                is_data = r_data.match(line)
-                if is_data:
-                    data = is_data.groupdict()
-                    if r_quotes.match(data["value"]):
-                        data["value"] = data["value"][1:-1]
-                    if data["orig_key"] == "AWSAccessKeyId" \
-                       or data["orig_key"] == "aws_access_key_id":
-                        data["key"] = "access_key"
-                    elif data["orig_key"]=="AWSSecretKey" \
```
fviard: I have found a reference about that:
https://blog.csanchez.org/2011/05/
Looks like it used to be the case up to 2011 at least.
Its a little bit annoying to have to take care of that, but that would probably be better to still handle that case.
Note: a warning might still be issued.
But users could have inherited config files or found old tutorial on internet and still be using that.
It is bad to break user existing setups just for updates when it is not strictly necessary.

Trane9991:
Hey @fviard,
your comments are very appreciated, originally I'm not the Python developer so could miss some obvious things :)

I pushed the fixes for all your comments, had to refactor the func to support the legacy options without much uglinesses and repetition in the code.

fviard:
Thank you for your reactivity anyway.
Looks good, i will give it a try tomorrow and merge it.

Frantch:
Consider adding credentials refresh. 

Imagine one wants to sync a very large data set and provides a profile that is valid for 60minutes. The sync will fail after 60 min and the command needs to be re run, this is very slow since it will have to re enumerate the full bucket

However if for example evert 35 minutes a background process update the credentials file every 45 minutes let's say, s3cmd when it fails for a token expires reason, it should retry loading the credentials file.

Trane9991:
@Frantch 
AFAIK `aws_access_key_id` and `aws_secret_access_key` keys that are provided for the AWS IAM *User* are entirely static and do not expire ever(automatically).

The ones you are talking about is the credentials provided by AWS IAM *Role* and read from ec2 instance metadata. This type of credentials not covered in this PR since they are already supported.

fviard:
@Frantch As said by Trane9991, this patch is not related to the same credentials.
For iam, I think that it will not expire because sadly credentials are refreshed every requests.
That is not optimal for performances but at least you should not encounter expiration issues.
It is something that I have on my todo list for improvements.

alexanderkjeldaas:
@Trane9991 I don't understand this:

> The ones you are talking about is the credentials provided by AWS IAM Role and read from ec2 instance metadata. This type of credentials not covered in this PR since they are already supported.

AWS IAM Roles look like this and I don't think they are supported.  Is `$AWS_PROFILE` really supported?

```
[profile bifrost]
role_arn = arn:aws:iam::<acct>:role/my-iam-role
source_profile = default
```

Trane9991:
@alexanderkjeldaas 
IAM roles don't work this way. You can't simply provide the role ARN and use it.

The role needs to be attached to the EC2 instance, then tool, that runs on this instance my use the role (get the credentials from Instance Metadata and user them). See [Using an IAM Role to Grant Permissions to Applications Running on Amazon EC2 Instances](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2.html).

Technically, what you want, could be achieved by [Assume Role mechanism](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_permissions-to-switch.html) and it is not supported AFAIK (at least not by this PR).

PR #968, opened 2018-04-30
Avoid distutils usage at runtime

LocutusOfBorg:
This avoids the distutils need as runtime dependency

### DIFF
--- a/s3cmd
+++ b/s3cmd
@@ -57,7 +57,12 @@ import tempfile
 from copy import copy
 from optparse import OptionParser, Option, OptionValueError, IndentedHelpFormatter
 from logging import debug, info, warning, error
-from shutil import which
+
+try:
+    from shutil import which
+except ImportError:
+    # python2 fallback code
+    from distutils.spawn import find_executable as which
 
 from ssl import SSLError
 import io
--- a/s3cmd
+++ b/s3cmd
@@ -57,7 +57,7 @@ import tempfile
 from copy import copy
 from optparse import OptionParser, Option, OptionValueError, IndentedHelpFormatter
 from logging import debug, info, warning, error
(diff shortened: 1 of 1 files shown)


fviard:
Thank you for the PR, in the theory i'm in favor to remove the dependency on distutils, but "which" is not available in python2 (2.6, 2.7) so I can't accept the PR as is or s3cmd would be broken for python 2.
Maybe, you can simply add a fallback to find_executable on ImportError ?

To be noted, but I don't know if there could be a chance that this affect an user, there is a subtle difference between find_package and which based on the following link:
https://stackoverflow.com/questions/9877462/is-there-a-python-equivalent-to-the-which-command
>  Keep in mind that unlike which, find_executable does not actually check that the result is marked as executable.

LocutusOfBorg:
Hello!

>Thank you for the PR, in the theory i'm in favor to remove the dependency on distutils, but "which" is >not available in python2 (2.6, 2.7) so I can't accept the PR as is or s3cmd would be broken for python >2.

I see..
>Maybe, you can simply add a fallback to find_executable on ImportError ?

>To be noted, but I don't know if there could be a chance that this affect an user, but there is a subtle >difference between find_package and "which":

I would say that which is superior then :)

LocutusOfBorg:
This should be better now...

fviard:
Thank you.
I would probably look picky, but can you change it again to only have imports at the top of the file?
I think that it would be ok to import which as "find_executable" or the opposite to use a single call in the body itself.

LocutusOfBorg:
Oh... it was so easy... I tried to remember a good way to do it, and now I got it! thanks for the help!

fviard:
Perfect, merged, thank you!

Issue #956, opened 2018-02-13
ls output misaligned (formatting issue)


vpresnyak:
Hello,

I've had some issues with the output of the `ls` command. It seems to misalign files, which makes them hard to read. This started with the upgrade to py3 and v2.0.1 - previous versions of s3cmd didn't seem to have this issue.

```
$ s3cmd ls s3://my_bucket/files/2017/

2017-03-10 17:29 14069483520   s3://my_bucket/files/2017/file1.tar
2017-03-10 17:29 32289597440   s3://my_bucket/files/2017/file2.tar
2017-03-15 16:13 283443200   s3://my_bucket/files/2017/file3.tar
2017-03-10 16:51 9882787840   s3://my_bucket/files/2017/file4.tar
2017-10-23 21:15 66645022720   s3://my_bucket/files/2017/file5.tar
2017-08-04 22:03 229539840   s3://my_bucket/files/2017/file6.tar
2017-03-10 16:48 12873990656   s3://my_bucket/files/2017/file7.tar
```
oddly, adding the `--human-readable-sizes` parameter only makes things worse. It doesn't look like it's rounding correctly.

```
$ s3cmd ls -H s3://my_bucket/files/2017/

2017-03-10 17:29 13.103227615356445G  s3://my_bucket/files/2017/file1.tar
2017-03-10 17:29 30.072031021118164G  s3://my_bucket/files/2017/file2.tar
2017-03-15 16:13  270.3125M  s3://my_bucket/files/2017/file3.tar
2017-03-10 16:51 9.204063415527344G  s3://my_bucket/files/2017/file4.tar
2017-10-23 21:15 62.06801414489746G  s3://my_bucket/files/2017/file5.tar
2017-08-04 22:03 218.90625M  s3://my_bucket/files/2017/file6.tar
2017-03-10 16:48 11.989838123321533G  s3://my_bucket/files/2017/file7.tar
```

Thanks!

mouchar:
The first issue (misaligned columns) appears with older Python and older s3cmd as well.
It is caused by too narrow column size for object size (9 characters) as defined in format_string `u"%(timestamp)16s %(size)9s%(coeff)1s  %(uri)s"`
As big file sizes are quite common these days, format might be extended at the expense of wider rows.

Problem with human-readable-sizes is definitely caused by the change how division is made by Python.
While Python 2 returns integer, Python 3 returns float. Simple wrapping with int() could improve the output and get rid of those ugly floats.

[...]

PR #986, opened 2018-07-17
Fix for proper handling of int and float file sizes

vpresnyak:
Fixing proper handling of file sizes. Will correctly return floating point or integer as appropriate.

Previous version appeared to rely on integer division as implemented in py2 and would return a float regardless of flating_point argument in py3.

### DIFF
--- a/S3/Utils.py
+++ b/S3/Utils.py
@@ -192,16 +192,16 @@ def dateRFC822toUnix(date):
 __all__.append("dateRFC822toUnix")
 
 def formatSize(size, human_readable = False, floating_point = False):
-    size = floating_point and float(size) or int(size)
+    size = int(size)
     if human_readable:
         coeffs = ['k', 'M', 'G', 'T']
         coeff = ""
         while size > 2048:
             size /= 1024
             coeff = coeffs.pop(0)
-        return (size, coeff)
+        return (float(size) if floating_point else int(size), coeff)
     else:
-        return (size, "")
+        return (float(size) if floating_point else int(size), "")
 __all__.append("formatSize")
 
 def formatDateTime(s3timestamp):


vpresnyak:
Hmm, I'm not clear on why this is failing testing. on py2.6 and py3.5 specifically with a `ConnectionRefusedError: [Errno 111] Connection refused`.

Would appreciate insights!

fviard:
I think that it was just a cache issue with CI.
Thanks for the PR that looks like to make sense, i will have a look at it.

vpresnyak:
I believe that this resolves #956. At least in my testing!

fviard:
I have merged you PR, thank you.
But for your info, it is breaking floating point on py2.
Because there it needs to have a "float" in input before the division to give a float result.
I will fix that don't worry.

vpresnyak:
Thanks for fixing!

That's odd, that file imports division behavior from py3. It wasn't breaking float division in py2 when I tested my version locally. I wonder if that behavior is system-dependent somehow.

fviard:
the difference is that python 3 always return float if needed even if both arguments of a division were int.
At the opposite of python2.

When I mean broken, i mean that if python 3 would have returned 96.7, python 2 was returning 96.0.

vpresnyak:
That's not accurate in this case. Consider the two division operations below in py2 and py3 that are run with the same imports as s3cmd/Utils.py. This demonstrates that once your import division from future, the division operation functions identically in py2 and py3.

```
Python 3.6.3 |Anaconda custom (64-bit)| (default, Oct 13 2017, 12:02:49)
[GCC 7.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from __future__ import absolute_import, division  # imports from line 9 of Utils.py
>>> print(42/11)
3.8181818181818183
```
```
$ python
Python 2.7.12 |Anaconda custom (64-bit)| (default, Jul  2 2016, 17:42:40)
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from __future__ import absolute_import, division
>>> print(42/11)
3.81818181818
```

fviard:
Indeed you are right, I forgot the __future__ import in the code and tested in an interpreter directly.
Sorry for error ^^

PR #995, opened 2018-08-07
Enable AWS cred file to be read as long as .s3cfg auth fields are nonexistent

jma127:
This commit allows the credentials to be read from ~/.aws/credentials
or user-specified (via envvar) file if .s3cfg exists but does not have
authentication options.

The previous behavior is that s3cmd will *not* read credentials from the
above path as long as .s3cfg exists (even if it contains no
authentication options).

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -24,7 +24,7 @@ except ImportError:
     import http.client as httplib
 import locale
 
-try: 
+try:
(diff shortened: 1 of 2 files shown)


jma127:
OK, this should be in a semi-decent state now. Please let me know if any changes are needed.

PR #1012, opened 2018-10-22
Add extra headers to all requests, if they exist

leondutoit:
At the moment, adding custom headers only works for simple put operations. With this change, one can use custom headers in any s3cmd operation.


patthiel:
I am also interested in seeing this change merged. I can add a test if needed

Issue #1006, opened 2018-09-30
TypeError: Unicode-objects must be encoded before hashing in `s3cmd sync --cf-invalidate`


benkuhn:
Thanks for making s3cmd, it's super useful!

I recently tried to use it with python3 and got the following traceback when using the `--cf-invalidate` flag:

```
Traceback (most recent call last):
  File "/Users/ben/.virtualenvs/bknet/bin/s3cmd", line 3092, in <module>
    rc = main()
  File "/Users/ben/.virtualenvs/bknet/bin/s3cmd", line 3001, in main
    rc = cmd_func(args)
  File "/Users/ben/.virtualenvs/bknet/bin/s3cmd", line 1885, in cmd_sync
    return cmd_sync_local2remote(args)
  File "/Users/ben/.virtualenvs/bknet/bin/s3cmd", line 1864, in cmd_sync_local2remote
    _invalidate_on_cf(destination_base_uri)
  File "/Users/ben/.virtualenvs/bknet/bin/s3cmd", line 1834, in _invalidate_on_cf
    results = cf.InvalidateObjects(destination_base_uri, uploaded_objects_list, default_index_file, cfg.invalidate_default_index_on_cf, cfg.invalidate_default_index_root_on_cf)
  File "/Users/ben/.virtualenvs/bknet/lib/python3.5/site-packages/S3/CloudFront.py", line 452, in InvalidateObjects
    cfuris = self.get_dist_name_for_bucket(uri)
  File "/Users/ben/.virtualenvs/bknet/lib/python3.5/site-packages/S3/CloudFront.py", line 587, in get_dist_name_for_bucket
    response = self.GetList()
  File "/Users/ben/.virtualenvs/bknet/lib/python3.5/site-packages/S3/CloudFront.py", line 323, in GetList
    response = self.send_request("GetList")
  File "/Users/ben/.virtualenvs/bknet/lib/python3.5/site-packages/S3/CloudFront.py", line 506, in send_request
    request = self.create_request(operation, dist_id, request_id, headers)
  File "/Users/ben/.virtualenvs/bknet/lib/python3.5/site-packages/S3/CloudFront.py", line 555, in create_request
    signature = self.sign_request(headers)
  File "/Users/ben/.virtualenvs/bknet/lib/python3.5/site-packages/S3/CloudFront.py", line 567, in sign_request
    signature = sign_string_v2(string_to_sign)
[...]

PR #1013, opened 2018-11-03
Resolve binary vs unicode encodings in py3.7

rgant:
This patch addresses #1006.

Hard for me to say if the changes I've made will break other versions of python. This works for 3.7.


fviard:
@rgant thank you very much for having the reported this issue and tried a PR.
I did push different fixes that will better fit python2 support and avoid breaking something else, but your PR was very helpful to provide me some insight with into this issue.

squarelab-slee:
Is fixed version released?

Issue #847, opened 2017-03-22
UnicodeDecodeError


baiyanchun:
I used s3cmd command (the latest s3cmd code I use) to put a file named in Chinese, when I use s3cmd info bucket, An unexpected error has occurred. The following is my operation process and result.

1 Create a bucket
s3cmd mb s3://bucket
Bucket 's3://bucket/' created

2 Create a file named in Chinese
touch 中文.txt

3 Put file to s3://bucket
s3cmd put 中文.txt s3://bucket
upload: '中文.txt' -> 's3://bucket/中文.txt'  [1 of 1]
0 of 0     0% in    0s     0.00 B/s  done

4 Ls this file
s3cmd ls s3://bucket
2017-03-21 02:55         0   s3://bucket/中文.txt

5 Get bucket info
s3cmd info s3://bucket

The error message:
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

Invoked as: /usr/local/bin/s3cmd info s3://bucket
Problem: UnicodeDecodeError: 'ascii' codec can't decode byte 0xe4 in position 496: ordinal not in range(128)
S3cmd:   1.6.1
python:   2.7.6 (default, Oct 26 2016, 20:30:19) 
[GCC 4.8.4]
environment LANG=en_US.UTF-8

Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/s3cmd-1.6.1-py2.7.egg/EGG-INFO/scripts/s3cmd", line 2996, in <module>
  File "/usr/local/lib/python2.7/dist-packages/s3cmd-1.6.1-py2.7.egg/EGG-INFO/scripts/s3cmd", line 2915, in main
  File "/usr/local/lib/python2.7/dist-packages/s3cmd-1.6.1-py2.7.egg/EGG-INFO/scripts/s3cmd", line 918, in cmd_info
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe4 in position 496: ordinal not in range(128)

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

baiyanchun:
the detail error log.
[s3cmd.txt](https://github.com/s3tools/s3cmd/files/860924/s3cmd.txt)

mdomsch:
The next output line should be a policy.  By chance does your policy have
unicode characters in it?  That's where it's failing.

fviard:
[...]

PR #1021, opened 2018-12-06
Return decoded response in get_policy and get_cors

offshore:
This patch allows cmd_info to work properly in case of getting any unicode chars in reply to bucket Policy/CORS requests.
Fixes at least #847

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -966,7 +966,7 @@ class S3(object):
         request = self.create_request("BUCKET_LIST", bucket = uri.bucket(),
                                       uri_params = {'policy': None})
         response = self.send_request(request)
-        return response['data']
+        return decode_from_s3(response['data'])
 
     def set_policy(self, uri, policy):
         headers = SortedDict(ignore_case = True)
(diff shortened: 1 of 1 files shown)


fviard:
Thank you very much for this and sorry for the long delay.
This is a very good catch!
Note that I don't think that the fix will be related to the exact issue reported in #847 as the reporter was not calling anything related to get policy or cors.

offshore:
@fviard I'm glad it's finally resolved.

In fact I've noticed this whet tryied to make s3cmd talk to ceph's radosgw, and after some investigations it seems to me that 1) radosgw doesn't talk about bucket policy properly and 2) s3cmd is missing correct case processing code for policies (note: that was 3 months ago, so I can be imprecise).

I have another patch, which improves this situation, but **I'm unable to test it properly**; although if there's an interest, I can submit another PR (that will require proper thru-looking before being merged) -- or just feel free to drop me an email, I'll share it with you.

fviard:
@offshore don't hesitate to submit the PR :-) I try to always look carefully before merging that's why it takes a little more time to merge than I would like.

If possible, try to put everything that you might have in the pull request comments, like backtraces, or reported error or debug logs. That helps.
Also, if you could be able to give a temporary access to a radosgw instance to test the issue/fix, that could help also :-)

PR #1032, opened 2019-02-25
Make connection pooling configurable

vainu-arto:
This adds a config file parameter and command line option to make connection reuse configurable.


fviard:
Thank you for your PR.
For the "close()" it looks like to be a good cleanup.
I think you can even go a little bit "further" by putting the "if conn:" directly in the "close" of ConnMan.

For having max conn configurable, I don't really see what kind of situation would require it.
Would you mind to share your use case that required it?

vainu-arto:
For my use case the connection pooling simply does not work. S3 will always timeout the connection before it can be retrieved from the pool and reused. As far as I understand, the error retrieved when then trying to use the dead connection leads to s3cmd throttling itself for no reason.

fviard:
Sorry for the long delay to reply to you.
I see, that's the kind of thing that I was thinking.
So, what would you think about adding an option like "connection_pooling" (boolean value true/false) instead of a configurable counter?
I guess that it should solve the issue and might be less confusing for a standard user.

vainu-arto:
@fviard That would also work for my use case. I was trying to make this generic so that it would also work for possible other use cases.

vainu-arto:
@fviard Sorry about the very long delay, I got stuck doing something completely different. I have pushed a new version implementing the connection_pooling boolean instead, let me know if there are other changes you'd like to see in this.

fviard:
Took me a year to merge this, but it was a very good contribution from vainu-arto!

Issue #666, opened 2015-11-20
"signurl" is ignoring "use_https" and other configuration settings


jamshid:
Seems the url generated by `signurl` should use the endpoint style described in `~/.s3cfg`. In particular:
* use `https` if `use_https = True` 
* put the bucket name in the path, not the domain/endpoint, if `host_bucket == host_base`

```
$ s3cmd --version   # installed with "brew install --HEAD s3cmd"
s3cmd version 1.6.0+
$ s3cmd signurl s3://mybucket/file.mp4 +3600
http://mybucket.storage.example.com/file.mp4?AWSAccessKeyId=1e71169c9ab10b293bda2b454bf20c35&Expires=1448001252&Signature=xcJUwTgRq5xSvTXp3K3zI3v%2Fq7g%3D
```
Expected:
```
https://storage.example.com/mybucket/file.mp4?AWSAccessKeyId=1e71169c9ab10b293bda2b454bf20c35&Expires=1448001252&Signature=xcJUwTgRq5xSvTXp3K3zI3v%2Fq7g%3D
```
Actual requests work fine: s3cmd uses `https://storage.example.com`. I'm using an S3-compatible service, here's my `~/.s3cfg`:
```
[default]
access_key = 1e71169c9ab10b293bda2b454bf20c35
secret_key = secret
use_https = True
host_base = storage.example.com
# Using first format to force "bucket in path not Host:"
host_bucket = storage.example.com
#host_bucket = %(bucket)s.storage.example.com
```

mdomsch:
The SSL certificates offered by AWS S3 (and other public cloud providers)
won't match your DNS hostname (either DNS-style bucket names, or CNAMEd
buckets).  That then leaves it up to clients to determine if they want to
allow the https connection, or not.  Recent python, for example, will
reject it.  So I don't think there's a huge benefit to s3cmd replying with
'https' on the front.  If in your situation you know http should be https
and will be accepted, you can always replace http with https in the
returned string after s3cmd is done.

I'm open to a better solution, but simply honoring use_https (which is
really for s3cmd's connections to S3, not other clients downstream of us -
this is an overload of the option) doesn't seem like the right solution.

Thanks,
Matt

zokier:
Would a configuration option be acceptable solution here? For example `signurl_base = http://%(bucket)s.%(host_base)s` as default which then could be changed to `signurl_base = https://%(host_base)s/%(bucket)s` to match @jamshid needs.

[...]

PR #1033, opened 2019-02-28
Configurable public URL protocol

junousi:
If public_url_use_https is true, prefix public URLs with https. Otherwise use
http as before.

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -205,6 +205,7 @@ class Config(object):
     # Maximum sleep duration for throtte / limitrate.
     # s3 will timeout if a request/transfer is stuck for more than a short time
     throttle_max = 100
+    public_url_use_https = False
 
     ## Creating a singleton
     def __new__(self, configfile = None, access_key=None, secret_key=None, access_token=None):
--- a/S3/S3Uri.py
+++ b/S3/S3Uri.py
@@ -91,7 +91,7 @@ class S3UriS3(S3Uri):
 
     def public_url(self):
         public_url_protocol = "http"
-        if Config.Config().use_https:
+        if Config.Config().public_url_use_https:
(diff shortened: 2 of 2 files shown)


fviard:
Hi, thank you for your contribution.
The question come back regularly, but having https as a default public url is creating a lot of issues.
See: https://github.com/s3tools/s3cmd/issues/666
The https for connecting to aws s3, is not the same as the public link serve a content.

If you would be willing to rework your patch, I think that it would be good to finally fix this issue by creating a dedicated config option for that (but that can't be enabled by default)

junousi:
Hi, sure thing!

I'm a bit unsure what to rework as public_url() seems to be mostly feeding output() or info() and not plugging into other s3cmd functionality, so would need a pointer for what else to change. In any case a simple config parameter separate from use_https is now amended.

junousi:
Also some background. We're running an object storage service with CephRGW/S3 open to the world only over https/443, hence any form of configurability for changing the public URL prefix from http to https would be a plus.

fviard:
Merged, thank you very much!

fviard:
In fact I just realized that we had "signurl_use_https".
signurl and public_url are basically the same kind of thing and if someone might accept one, the other one is probably acceptable for him.
signurl_use_https would have been a bad name if it was acting on public_uri also.

So probably, the right config should have been something like "url_use_https", or "use_https_for_url" and be shared for the 2 usages. But it would be annoying to change the name of a config that is already there for a few years :-s

junousi:
> But it would be annoying to change the name of a config that is already there for a few years :-s

Agreed. In time perhaps both parameters can be superseded with the common one, but I think it's better to do this small incremental change first. Cheers!

PR #1034, opened 2019-03-21
Add missing return

Jamstah:
Batch delete fails to return a response object so callers cannot detect failures

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -717,7 +717,7 @@ class S3(object):
     def object_batch_delete(self, remote_list):
         """ Batch delete given a remote_list """
         uris = [remote_list[item]['object_uri_str'] for item in remote_list]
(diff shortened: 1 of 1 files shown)


fviard:
Hi,
I know that it is a little bit a long time ago, but this would not have been enough to change the behavior of anything as the caller is not looking at the reply anyway.
Can you tell me a little bit more about the context or the issue itself that you were encountering?
Thanks

fviard:
Merged thank you.

I got no reply, but I guess that it was requested only to be able to be used when "S3" is used as a lib for another program.
The change is harmless and makes the code more consistent.

Jamstah:
Gosh, I can't remember what I was doing with it at the time, but thanks for the merge :)

Like you say, I was probably using it as a library.

PR #1044, opened 2019-06-13
Fix Python 3 compatibility

thomasgoirand:
Fix Python 3 compatibility

### DIFF
--- a/S3/ACL.py
+++ b/S3/ACL.py
@@ -11,6 +11,9 @@ from __future__ import absolute_import, print_function
 import sys
 from .Utils import getTreeFromXml, deunicodise, encode_to_s3, decode_from_s3
 
+import six
+
+
 try:
     import xml.etree.ElementTree as ET
 except ImportError:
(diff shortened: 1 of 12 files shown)


fviard:
Hum, I'm sorry but I can't merge that.
I don't see what this brings except subtly breaking a few things and creating an useless dependency on six?

Can you tell me what is the compatibility bug that your tried to fix?
Because from what I see, this PR just change code that is working for that purpose.

S3/ACL.py
```diff
@@ -207,7 +210,7 @@ def __unicode__(self):
     def __str__(self):
         if PY3:
             # Return unicode
-            return ET.tostring(self.get_printable_tree(), encoding="unicode")
+            return ET.tostring(self.get_printable_tree(), encoding="six.text_type")
```
adamchainz: THis is an incorrect find/replace

fviard:
As stated before, doesn't look like to solve any specific bug or have any effect.
Please re-open if you can reply to my previous comment. Thanks

PR #1051, opened 2019-07-18
Allow setting SSL client authentication

chazer:


fviard:
Hi, I'm sorry to have a look at that a little bit late.
I think that I see how it can be useful, but can you tell me more about the ssl client authentication.
Is there any public service that support that? Does aws s3 support it?
What is the purpose? the key authentication is not enough?

chazer:
Hi! Nothing wrong, I did not wait and made a fork)
Certificate-based authentication is a more flexible mechanism for creating keys with a limited lifetime, with the possibility of revocation. Any TCP protocol can be wrapped in an SSL/TLS tunnel.
I'm not sure AWS provides such a service, but you can always put proxies in front of AWS or another cloud provider.
The s3cmd can be used not only with AWS, but also with alternative programs that provide the S3 protocol, such as Zenko Cloudserver, s3proxy. In my work, I use s3proxy to convert various cloud storage protocols into an S3 protocol (like OpenStack Swift), and there is no way to flexibly configure key authorization and permissions.

fviard:
Thank you very much for your explanation.
Now I understand and it makes sense.
I think that it is a great addition.
I would like to merge it, but I think that there are a little few things that could improve this PR.
See inline messages for my comments.

Would it be possible for you to have a look at the suggested changes (and eventually rebase) in a short term? 
Otherwise I will merge this and modify it myself after.
The purpose is to be able to put that in the next release that I will do very soon.
Thanks

S3/ConnMan.py
```diff
@@ -73,9 +86,20 @@ def _ssl_context():
         cafile = cfg.ca_certs_file
         if cafile == "":
             cafile = None
+        certfile = cfg.ssl_cert_file
+        if certfile == "":
+            certfile = None
+        keyfile = cfg.ssl_key_file
+        if keyfile == "":
+            keyfile = None
+
```
fviard: For the previous blocks, including cafile, you can simplify it:
ex.:
> cafile = cfg.ca_certs_file or None

(There is a lot of non really clean code all other s3cmd, but it is legacy, and not a code style to follow :-) )

S3/Config.py
```diff
@@ -149,6 +149,8 @@ class Config(object):
     gpg_decrypt = u"%(gpg_command)s -d --verbose --no-use-agent --batch --yes --passphrase-fd %(passphrase_fd)s -o %(output_file)s %(input_file)s"
     use_https = True
     ca_certs_file = u""
+    ssl_key_file = u""
+    ssl_cert_file = u""
```
fviard: I think that maybe it would be more easy for a "non expert" to understand these configs and cmd arguments if you rename them to:
> ssl_client_key_file
> ssl_client_cert_file

S3/ConnMan.py
```diff
@@ -73,9 +86,20 @@ def _ssl_context():
...
         debug(u"Using ca_certs_file %s", cafile)
+        debug(u"Using ssl_cert_file %s", certfile)
+        debug(u"Using ssl_key_file %s", keyfile)
 
-        if cfg.check_ssl_certificate:
+        if keyfile is not None and certfile is not None:
```
fviard: I think that you should modify the condition to:
> if certifle is not None:

If I understand well, it is possible to have just a certifle and not a keyfile (because the private keys are embedded into the certifle), but the opposite is not true.
And the rest of the code should be still ok even if "keyfile" value is None.

s3cmd
```diff
@@ -2755,6 +2755,8 @@ def main():
     optparser.add_option(      "--cache-file", dest="cache_file", action="store", default="",  metavar="FILE", help="Cache FILE containing local source MD5 values")
     optparser.add_option("-q", "--quiet", dest="quiet", action="store_true", default=False, help="Silence output on stdout")
     optparser.add_option(      "--ca-certs", dest="ca_certs_file", action="store", default=None, help="Path to SSL CA certificate FILE (instead of system default)")
+    optparser.add_option(      "--ssl-cert", dest="ssl_cert_file", action="store", default=None, help="Path to client SSL certificate FILE")
+    optparser.add_option(      "--ssl-key", dest="ssl_key_file", action="store", default=None, help="Path to client SSL certificate key FILE")
```
fviard: See previous comment about the option names

s3cmd.1
```diff
@@ -539,6 +539,12 @@ Silence output on stdout
 Path to SSL CA certificate FILE (instead of system
 default)
 .TP
+\fB\-\-ssl\-cert\fR=CRT_FILE
+Path to client SSL certificate CRT_FILE
+.TP
+\fB\-\-ssl\-key\fR=KEY_FILE
+Path to client SSL certificate key KEY_FILE
+.TP
```
fviard: I think that maybe it will help the understanding of novice users to modify the help messages to:
> Path to client own SSL certificate CRT_FILE
> Path to client own SSL certificate private key KEY_FILE

S3/ConnMan.py
chazer: Yeah, you were right!

s3cmd
chazer: No, names should not be made long. But in the description of the parameters I will make changes
For cURL, they are simply called "key" and "cert".

s3cmd.1
chazer: OK

fviard:
Merged, thank you very much for this new feature!

PR #1052, opened 2019-08-04
Enable cfinval command

erseco:
Enabled the commented `cfinval` command to allow invalidating specific cloudfront files


fviard:
Thanks for the contribution.
The 1 million dollar question before merging this is "why was it commented out in the first place?"
Sadly, git history does not give a clue about that.

fviard:
cfinval enabled with a recent push to MASTER. See #1256

To reply the previous question, the command was not enabled because the code to do the feature was not existing...

Issue #945, opened 2018-01-02
RuntimeError: dictionary changed size during iteration


LocutusOfBorg:
Hello, I'm forwarding from bugs.debian.org/885914

```
Package: s3cmd
Version: 2.0.1-1
Severity: grave
Justification: renders package unusable

Dear Maintainer,

s3cmd put is failing with:

    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
        An unexpected error has occurred.
      Please try reproducing the error using
      the latest s3cmd code from the git master
      branch found at:
        https://github.com/s3tools/s3cmd
      and have a look at the known issues list:
        https://github.com/s3tools/s3cmd/wiki/Common-known-issues-and-their-solutions
      If the error persists, please report the
      following lines (removing any private
      info as necessary) to:
      s3tools-bugs@lists.sourceforge.net
   
   
    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
   
    Invoked as: /usr/bin/s3cmd --ssl --config=/tmp/tmp.akNQcIMsdc --acl-private put --cache-file=«path»/.s3cmd.cache --encrypt «files...» s3://«bucket»/
    Problem: <class 'RuntimeError: dictionary changed size during iteration
    S3cmd:  2.0.1
    python:  3.6.4 (default, Dec 19 2017, 14:09:48)
    [GCC 7.2.0]
    environment LANG=en_GB.UTF-8

    Traceback (most recent call last):
      File "/usr/bin/s3cmd", line 3073, in <module>
        rc = main()
      File "/usr/bin/s3cmd", line 2989, in main
        rc = cmd_func(args)
      File "/usr/bin/s3cmd", line 364, in cmd_object_put
        local_list, single_file_local, exclude_list, total_size_local = fetch_local_list(args, is_src = True)
      File "/usr/lib/python3/dist-packages/S3/FileLists.py", line 353, in fetch_local_list
        _maintain_cache(cache, local_list)
      File "/usr/lib/python3/dist-packages/S3/FileLists.py", line 311, in _maintain_cache
        cache.purge()
      File "/usr/lib/python3/dist-packages/S3/HashCache.py", line 49, in purge
        for i in self.inodes[d].keys():
    RuntimeError: dictionary changed size during iteration

/tmp/tmp.akNQcIMsdc contains:
    [default]
    access_key = «access key»
    acl_public = False
    bucket_location = EU
[...]

PR #1070, opened 2020-01-30
Fix for Issue #945 ("RuntimeError: dictionary changed size during iteration" when using cache-file)

NotTheDr01ds:
Fix for Issue #945 

Apologies for the delay, but here's the discussed fix for the issue.

Also added test cases to run-tests.py for the issue and fix.  It essentially:

* Modifies the first sync test to use a cache-file.  This in-and-of-itself should have no bearing on the success or failure of that particular test; it just sets things up for the following two steps.  The cache file is created in a new subdirectory of testsuite, so that it can be removed afterwards without impacting subsequent tests.
* Creates a file that didn't exist during the first sync.  This is also done in the new subdirectory of testsuite.
* Syncs again, confirming that the file was synced.
* Removes that file and does another sync.  This is the step that would have generated a condition causing the "dictionary changed size" error.  I confirmed that this test fails without the changes to HashCache.py, but passes with them.
* Finally, clean up the subdirectory which contained the cache file.

Tests were run (using the Docker framework I previously submitted in PR#1068) on:

* Python 2.7.17
* Python 3.7.6
* Python 3.8.1

And, of course, PR#1067 should also be merged in order to successfully run all test cases.  Those changes are not included in this PR.

### DIFF
--- a/S3/HashCache.py
+++ b/S3/HashCache.py
@@ -31,9 +31,9 @@ class HashCache(object):
         return d['md5']
 
     def mark_all_for_purge(self):
-        for d in self.inodes.keys():
-            for i in self.inodes[d].keys():
-                for c in self.inodes[d][i].keys():
+        for d in tuple(self.inodes):
+            for i in tuple(self.inodes[d]):
+                for c in tuple(self.inodes[d][i]):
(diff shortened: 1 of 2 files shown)


fviard:
Merged, thank you very much!

PR #1071, opened 2020-02-14
README: brew install and renamed to INSTALL.md

nitrocode:
Noticed that someone created a `brew` formula for this and thought it would be nice if it was documented somewhere in the `README.md` or the `INSTALL`. While looking at the `INSTALL`, I realized it wasn't easy to read so i renamed it to `INSTALL.md`


fviard:
Merged, thank you!

PR #1083, opened 2020-03-27
add support of aws server-side encryption with customer-provided keys

inspur-wyq:
Add support of aws server-side encryption with customer-provided encryption keys [https://docs.aws.amazon.com/AmazonS3/latest/dev/ServerSideEncryptionCustomerKeys.html](url)

We can use option --server-side-encryption-customer-key=CUSTOMER_KEY to use SSE-C when putting objects or get the SSE-C object.


fviard:
Sorry that I have not replied earlier, but interesting/needed feature.
I will have a look at merging that.

salanki:
Will this get merged?

PR #1100, opened 2020-05-09
Support HTTP-Chunked

giorgiozoppi:
During the test of our S3 solution, we have experienced an error that it is due mishandling to HTTP chunked mode. Simply s3cmd doesn't support chunked mode by default. This patch is just a couple line of codes to support HTTP-Chunk mode.
Chunked transfer encoding is a streaming data transfer mechanism available in version 1.1 of the Hypertext Transfer Protocol (HTTP). In chunked transfer encoding, the data stream is divided into a series of non-overlapping "chunks". The chunks are sent out and received independently of one another. No knowledge of the data stream outside the currently-being-processed chunk is necessary for both the sender and the receiver at any given time.

	modified:   S3/Exceptions.py
	modified:   S3/MultiPart.py
	modified:   S3/S3.py

### DIFF
--- a/S3/Exceptions.py
+++ b/S3/Exceptions.py
@@ -99,7 +99,6 @@ class S3Error (S3Exception):
         if error_msg:
             retval += (u": %s" % error_msg)
         return retval
-
     def get_error_code(self):
         if self.status in [301, 307]:
             return ExitCodes.EX_SERVERMOVED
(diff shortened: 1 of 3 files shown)


giorgiozoppi:
Please dont' merge or stash, i am on it.

giorgiozoppi:
We close this.

PR #1118, opened 2020-06-27
[#1117] Improve speed when running on EC2 instance

### LINKED ISSUE #1117: iam role (access token) is refreshed for every s3 request
iam role (access token) is refreshed for every s3 request , and so, that can result in a lot slower operations as a request for credential has to be done each time, even if access token is not expired.

A nice fix would be to take into account the expiration date of the token, to only refresh it when it is needed.

Patouche:
Closes #1117 .

Fix http call on role endpoint to retrieve the token each time the role_refresh in invoke

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -17,6 +17,26 @@ import sys
 import json
 from . import Progress
 from .SortedDict import SortedDict
+import datetime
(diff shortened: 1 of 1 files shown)


fviard:
Merged :-)

PR #1120, opened 2020-06-30
Update wiki URL in error message

anbnyc:
The wiki title and therefore its URL changed a few months ago, but the error message still points to the old URL.


fviard:
Good catch, merged, thank you for the PR!

Issue #1128, opened 2020-08-27
HTTP Proxy Authentication


islab1920:
Does s3cmd support HTTP proxy authentication? Our environment requires users to authorize with ID and password.

```
$ s3cmd ls s3://foobar/
ERROR: Error parsing xml: mismatched tag: line 65, column 2
ERROR: b'TRIMMED! GATEWAY RESPONSE'
ERROR: S3 error: 407 (Proxy Authentication Required)
```

### ~/.s3cfg
```
proxy_host = proxy.foobar.com
proxy_port = 8080
```

fviard:
As of today, http proxy authentication is not supported.

It should be trivial to add, but I don't have the proxy setup to test.
If I try to add this feature quickly, would it be possible for you to test it?

mikoim:
@islab1920 I've implemented proxy authentication in #1129. Please give it a try.

fviard:
@mikoim Thank you for the PR. I'm sorry I was in a trip and had not yet the time to reply to you.
As is, I don't think that it will work yet, but I will give you the suggestions to fix it in the PR.

PR #1129, opened 2020-09-03
Add support for authenticated proxies

mikoim:
This PR allows s3cmd to use an authenticated proxy server.
#1128

## TODO
- [x] use_https = True -> Set ``Proxy-Authenticate`` header via HTTPConnection.set_tunnel()
- [ ] use_https = False -> Need help


fviard:
Thank you for your contribution!
Here are my suggestions for a few improvements:
- don't use encode/decode but, "deunicodise_s" and "unicodise_s" from Utils.
And so, there will be no need to specify the encoding.
(as there is no issue to decode ascii with utf-8).

- Instead of creating new config entries for proxy_username, proxy_password, it would be better to reuse the proxy_host entry so, configuration is not changed.
the id would be to support a syntax like:
proxy_host = login:pass@myhost
For that, nothing special to do, just at the time of usage, you have 2 choices:
* Just detect the @ in proxy_host and split by @ and : to extract login/pass/host.
* Use urlparse to extract the info.

- For use_https = False, it looks like that there is no nice way to do that, but I'm still thinking about it.

mikoim:
@fviard Thank you for your advice. For backward-compatibility, ```proxy_host``` and ```proxy_port``` should be kept, but I guess that the format "login:pass@myhost" is not common. How about introducing new parameter like a ```proxy_url```, which supports "schema://username:password@host:port/"?

s3cmd
```diff
@@ -68,6 +68,12 @@ except ImportError:
     # python2 fallback code
     from distutils.spawn import find_executable as which
 
+try:
+    # python 3 support
+    from urlparse import urlparse
+except ImportError:
+    from urllib.parse import urlparse
```
fviard, line 75: It is a detail, but you can probably exchange the imports and have the python3 "from urlib.parse import urlparse" case first as it should now be the default case.

S3/ConnMan.py
```diff
@@ -229,14 +230,22 @@ def __init__(self, id, hostname, ssl, cfg):
                 self.c = httplib.HTTPConnection(self.hostname, self.port)
                 debug(u'non-proxied HTTPConnection(%s, %s)', self.hostname, self.port)
         else:
+            headers = {}
+            proxy_hostname = cfg.proxy_host
+            if '@' in cfg.proxy_host:
+                credential, proxy_hostname = cfg.proxy_host.split('@')
```
fviard, line 236: it is probably better to use split('@', 1) to avoid unexpected issues.

S3/ConnMan.py
```diff
@@ -229,14 +230,22 @@ def __init__(self, id, hostname, ssl, cfg):
...
+                # FIXME: Following line can't handle username or password including colon
+                proxy_username, proxy_password = credential.split(':')
```
fviard, line 238: Same a previous, split(':', 1)

fviard:
Thank your for your update and sorry for the long delay of my reply, sadly it was a very busy month :-s 

> @fviard Thank you for your advice. For backward-compatibility, proxy_host and proxy_port should be kept, but I guess that the format "login:pass@myhost" is not common. How about introducing new parameter like a proxy_url, which supports "schema://username:password@host:port/"?

Your point makes sense, but I'm a little bit reluctant to add a new option because of the confusion that it would induce.
(As I think that s3cmd is already a little bit scary as it has a lot of options)
And we would have still to support the proxy_host/proxy_port option for a long time for backward compatibility.

But something that could be done, I think, is to already accept url with that pattern in proxy_host.
Like, if there is a port in proxy_host, it overrides proxy_port.

PR #1137, opened 2020-10-02
fix compatibility with Python 3.9

ondrejbudai:
`getchildren()` method was removed from the `ElementTree` and `Element` classes in Python 3.9. This PR uses built-in `iter()` and `list()` methods instead of the removed one. This fix was suggested in the Python 3.9 release notes:

https://docs.python.org/3.9/whatsnew/3.9.html#removed

I verified this fix on Fedora 33 with Python 3.9.0rc2 by:

```
s3cmd --acl-public sync . s3://...
```

### DIFF
--- a/S3/Exceptions.py
+++ b/S3/Exceptions.py
@@ -127,7 +127,7 @@ class S3Error (S3Exception):
         if not error_node.tag == "Error":
             error_node = tree.find(".//Error")
         if error_node is not None:
-            for child in error_node.getchildren():
+            for child in error_node:
                 if child.text != "":
                     debug("ErrorXML: " + child.tag + ": " + repr(child.text))
                     info[child.tag] = child.text
--- a/S3/Utils.py
+++ b/S3/Utils.py
@@ -65,9 +65,9 @@ def parseNodes(nodes):
     retval = []
     for node in nodes:
         retval_item = {}
-        for child in node.getchildren():
+        for child in node:
             name = decode_from_s3(child.tag)
-            if child.getchildren():
+            if len(child):
                 retval_item[name] = parseNodes([child])
             else:
                 found_text = node.findtext(".//%s" % child.tag)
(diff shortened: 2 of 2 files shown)


S3/Exceptions.py
```diff
@@ -127,7 +127,7 @@ def parse_error_xml(tree):
         if not error_node.tag == "Error":
             error_node = tree.find(".//Error")
         if error_node is not None:
-            for child in error_node.getchildren():
+            for child in iter(error_node):
```
fviard, line 130: in the "for" loop, you can use directly error_node as it is iterable.
No need for iter() in that case.
Same thing with the for in your 2 other changes.

fviard:
Hi,
Thank you very much for your contribution.
I did not notice this change in python 3.9 and so this PR is very valuable.
If it is possible for you to fix your PR with the change that I suggested in my previous comment, it would be great and then I would be able to merge it.

ondrejbudai:
I force-pushed a fix - I incorporated your suggestion and I also used `len(child) > 0` instead of `list(child)` - I think that is more readable.

S3/Exceptions.py
ondrejbudai: Fixed

fviard:
Thank you for you read reactivity!
Good idea the 'if len(child) > 0' but can you just use 'if len(child)' instead please?
It is explicit enough and otherwise pylint might complain.

ondrejbudai:
Sure thing, done!

fviard:
Merged thanks!

ThiefMaster:
Could this please be released as 2.1.1?

KeithCu:
I'm running into this problem with every non-empty bucket in Linode. Can a new release be made soon?

leematos:
@KeithCu -- I had the same today which led me to pull master.zip, and if usr/bin/env python (symlink to python3) is available on the system, it can work. 

I agree though, A new release will be easier for more folks.

PR #1145, opened 2020-10-25
s3cmd: cleanup check_md5 handling a bit

xrmx:
Make the handling of the two opposite options symmetric.


fviard:
Merged, thank you!

PR #1147, opened 2020-10-29
Check self.access_token after set.

sungminoh:
Afaik, aws cli supports access token to be set in the credential file.
So I have a local cron job that updates access token in `.aws/credentials` and `.s3cfg` file.
By this change, s3cmd won't request a token if it is set in `.s3cfg` or `.aws/credentails` though it is not given by command line

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -273,7 +273,8 @@ class Config(object):
                 self.secret_key = secret_key
             if access_token:
                 self.access_token = access_token
-                # Do not refresh the IAM role when an access token is provided.
+            # Do not refresh the IAM role when an access token is provided.
+            if self.access_token:
                 self._access_token_refresh = False
 
             if len(self.access_key)==0:


S3/Config.py
```diff
@@ -273,7 +273,8 @@ def __init__(self, configfile = None, access_key=None, secret_key=None, access_t
                 self.secret_key = secret_key
             if access_token:
                 self.access_token = access_token
-                # Do not refresh the IAM role when an access token is provided.
+            # Do not refresh the IAM role when an access token is provided.
+            if self.access_token:
```
snosratiershad, line 277: It's basically equal to previous if statement

fviard: I don't think that it is the case if i'm not wrong.

Without this fix, if the access_token is set directly in the configuration file (and not set with command line or env var); then the access_token_refresh will not be disabled.

But, I'm wondering if there is not a side effect to this. Something like the goal was especially to not be stuck if an access token was written in the file. Especially for the case of a "configure" that would write the access token.

Issue #1151, opened 2020-11-17
Setting acl to private doesn't revoke public-write acl


orztt:
There is object  whoes ACL is public-read-write, which i used awscli setting.
Then i used `s3cmd setacl s3://object --acl-private` to set object acl to private. However, the object'acls  still have ` ACL: *anon*: WRITE ` item.
I can understand s3cmd only provides "acl-public", not "acl-public-read-write" for safety reasons. But i think "acl-private" should be consistent with awscli. 
So, i think `s3cmd setacl s3://objetct --acl-private` should revoke ` ACL: *anon*: WRITE ` item.

orztt:
There is my fix pr #1152

fviard:
Good idea, your PR was merged.

PR #1152, opened 2020-11-17
ACL: when setacl private, revoke public write acl

orztt:
fixes issuse #1151


fviard:
It is a little bit annoying that the command is not "symmetric", but it is probably good for security.
Merged, thanks!

PR #1155, opened 2020-11-23
Unite check for doing token refresh

### LINKED ISSUE #1153: s3cmd incorrectly tries to refresh when a token is provided through configuration file
When provided on command line or through environment variables, s3cmd correctly disables token refresh whereas if provided through the configuration file, it will try token refreshes causing pauses between each part transferred of a multipart object as well as log entries such as 

`WARNING: Could not refresh role`

pontus:
Move role refresh determination to a single check if a token
has been provided. This to make sure we disable token refresh if token is given through configuration file as well.

Fixes #1153 .

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -284,13 +284,15 @@ class Config(object):
                     self.secret_key = config_unicodise(env_secret_key)
                     if env_access_token:
                         self.access_token = config_unicodise(env_access_token)
-                else:
-                    self.role_config()
 
-            if self.access_token:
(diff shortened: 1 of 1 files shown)


fviard:
Hi, thank you for your contribution.

I might be wrong, but the access_token could be set during role_config, and in that case, we expect it to be refreshed.
With your change, after a role_config we will see a self.access_token and disable the refresh.
Or am i wrong?

pontus:
Ah, yes, seems like it. Apologies.

That particular issue should be addressed now.

PR #1160, opened 2020-12-04
Add link to install.md

siakaramalegos:
Adding a quick link to the install instructions since it was not obvious to me that I could just use brew before I downloaded the whole thing and found the install instructions. :)


fviard:
Merged, thank you!

PR #1165, opened 2020-12-12
Change info of sync command and --no-delete-removed option

FFKL:
I've updated descriptions of `sync` command and `--no-delete-removed` option and regenerated `s3cmd.1`.
Added remote to remote synchronization params for `sync` according to this code:
https://github.com/s3tools/s3cmd/blob/55597d16937085be3fc8894f23d958241a9aad9f/s3cmd#L1945-L1946
And `--no-delete-removed` can be used only for `sync`.


fviard:
Merged, thank you!

PR #1169, opened 2020-12-21
Require at least python 2.7

pinkavaj:
Tests are run only for Python 2.7, Python 2 itself is no longer supported.
It does not make much sense to support such a old version of languague.


fviard:
Thank you, I'm closing because for the moment I don't see the interest to restrict our support for older version if we can still support them.

PR #1171, opened 2020-12-23
Accelerate md5 checking of sync operation

yanghonggang:
- when sync objects from remote to local, store md5 into object file's 'user.s3cmd.md5' xattr
- when object file's md5 is first time calculated, store it into 'user.s3cmd.md5' xattr

When we need local file's md5, try to read it from 'user.s3cmd.md5' first(when '--use-md5-xattr'
is specified). Only calculate its md5 when 'user.s3cmd.md5' is not exist.

Signed-off-by: Yang Honggang <yanghonggang@kuaishou.com>

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -252,6 +252,7 @@ class Config(object):
     # seconds or more, but in real life, undocumented, it closes https conns
     # after around 6s of inactivity.
     connection_max_age = 5
+    use_md5_xattr = False
 
     ## Creating a singleton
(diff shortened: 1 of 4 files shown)


yanghonggang:
@fviard

fviard:
Sorry that I never replied to your PR.
This will not really work well, be quite dangerous in fact, because the file content could change, but the md5 inside xattribute will stay the same.
Also, we might prefer to not tinkle with the data that we are syncing.
In the end, you to avoid recalculating the md5 of data, you can use the "cache" feature that stores them inside a dedicated file.

PR #1198, opened 2021-06-30
Fixes s3cmd#1197 - add custom header for s3cmd mb

sebvajda:
In the Cloudian Hyperstore S3 software stack (www.cloudian.com) bucket creation can take an extra header to specify a storage policy id.

`s3cmd mb` silently ignores custom headers as specified with the `--add-header` option.

s3cmd and S3/S3.py updated to pass custom header to the S3 stack.

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -405,8 +405,11 @@ class S3(object):
         #debug(response)
         return response
 
-    def bucket_create(self, bucket, bucket_location = None):
+    def bucket_create(self, bucket, bucket_location = None, extra_headers = None):
         headers = SortedDict(ignore_case = True)
+        if extra_headers:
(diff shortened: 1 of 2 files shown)


sihara:
I also needed '--add-header' at the bucket creation and added codes myself. However, in the end, I found this commit which was exactly same of what I did. I totally agreed this patch and would proceed review to merge.

fviard:
Merged, thank you for your contribution!

Btw, if there are specific instructions to configure s3cmd to be used with Cloudian, don't hesitate to add the info to the following wiki page:
https://github.com/s3tools/s3cmd/wiki/How-to-configure-s3cmd-for-alternative-S3-compatible-services

PR #1199, opened 2021-07-07
Migrate CI tests to use github actions

arnaudj:
Use github actions for build and tests instead of TravisCI

Limitations:
- to bust the cached minio version, the cache key
  should be changed manually in the workflow file


fviard:
Merged, thank you :-)

arnaudj:
Np, thanks for the feedback during the testing phase :)
Cheers

Issue #1203, opened 2021-08-12
Wrong URL in restore request XML namespace


akete:
When using s3cmd with MinIO server (Version: 2021-08-05T22:01:19Z), restore object request results in `400 Bad Request error`:

 <pre>$ s3cmd --config s3cmd-local-minio.conf restore s3://bucket0/other/testFile
 ERROR: S3 error: 400 (MalformedXML): expected element <RestoreRequest> in name space http://s3.amazonaws.com/doc/2006-03-01/ but have http://s3.amazonaws.com/doc/2006-3-01</pre>

fviard:
Merged in Master.

PR #1204, opened 2021-08-12
Fixed RestoreRequest XML namespace URL (#1203)

akete:
Fixed RestoreRequest namespace according to S3 API: https://docs.aws.amazon.com/AmazonS3/latest/API/API_RestoreObject.html


fviard:
Merged, thank you very much!

Issue #352, opened 2014-06-29
Trying SSE-C


ahnkle:
I liked the idea of server side encryption with my keys, so I tried

``` bash
#!/bin/sh -ex
#
#

key=$(echo -n "key" | base64)
key_md5=$(echo -n "key" | md5sum | cut -b 1-32 | base64)

s3cmd \
  --add-header=x-amz-server-side-encryption-customer-algorithm:AES256 \
  --add-header=x-amz-server-side-encryptiony:"$key" \
  --add-header=x-amz-server-side-encryption-customer-key-MD5:"$key_md5" \
  $@
```

but I keep getting 

>  ERROR: S3 error: 403 (SignatureDoesNotMatch): The request signature we calculated does not match the signature you provided. Check your key and signing method.

Any ideas?

Example data values:

> base64 of "key": a2V5
> base64 of md5 of "key": M2M2ZTBiOGE5YzE1MjI0YTgyMjhiOWE5OGNhMTUzMWQK

mdomsch:
i'd be open to merging your branch. Care to rebase to current master and submit a pull request?
Thanks,
Matt

jrnt30:
@jheller Thanks for note about the binary vs. hex mode.  I was banging my head against a way for several hours until I stumbled upon this.

brice-laurencin:
This command should be in the documentation.

```
key=$(echo -n $secret | base64)
key_md5=$(echo -n $secret | openssl dgst -md5 -binary | base64)
```

I just spent literal days to find this.

mirdulvultr:
s3cmd put sample2.txt s3://publicstorage1245 --add-header="x-amz-server-side-encryption-customer-algorithm:AES256" --add-header="x-amz-server-side-encryption-customer-key:$key" --add-header="x-amz-server-side-encryption-customer-key-md5:$key_md5"

ran this command somehow put works flawlessly but when i try to run 

s3cmd get s3://publicstorage1245/sample2.txt -v -d --add-header="x-amz-server-side-encryption-customer-algorithm:AES256" --add-header="x-amz-server-side-encryption-customer-key:$key" --add-header="x-amz-server-side-encryption-customer-key-MD5:$key_md5"

it shows bad request error, does s3cmd not allow it ?

dimakond:
@mirdulvultr 

It looks like this parameter are working only during uploading objects.

`  --add-header=NAME:VALUE
                        Add a given HTTP header to the upload request. Can be
[...]

PR #1217, opened 2021-10-28
Add support for SSE-C encryption

noonedeadpunk:
Changes implement 2 new flags --sse-customer-key and
--sse-copy-source-customer-key that can be used by user to
provide a key for server side encryption.

Once these options are set extra headers are added to request
accordingly to SSE-C specification [1]

This PR squashes and rebases on current master changes
implemented by @jheller

[1] https://docs.aws.amazon.com/AmazonS3/latest/userguide/specifying-s3-c-encryption.html
Closes-Bug: #352

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,14 +34,21 @@ jobs:
         env:
           cache-name: cache-certgen
         with:
-          path: ~/cache
+          path: ~/.minio/certs
           key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ env.cache-revision }}
-      - name: Download minio and certgen on cache miss
-        if: steps.cache-minio.outputs.cache-hit != 'true' || steps.cache-certgen.outputs.cache-hit != 'true'
+      - name: Download minio on cache miss
+        if: steps.cache-minio.outputs.cache-hit != 'true'
         run: |
           mkdir -p ~/cache
           test ! -e ~/cache/minio && wget -O ~/cache/minio https://dl.minio.io/server/minio/release/linux-amd64/minio || echo "Minio already in cache"
-          test ! -e ~/cache/certgen && wget -O ~/cache/certgen https://github.com/minio/certgen/releases/download/v0.0.2/certgen-linux-amd64 || echo "Certgen already in cache"
+      - name: Download certgen on cache miss
+        if: steps.cache-certgen.outputs.cache-hit != 'true'
+        run: |
+          mkdir -p ~/.minio/certs
+          test ! -e ~/.minio/certs/certgen && wget -O ~/.minio/certs/certgen https://github.com/minio/certgen/releases/download/v0.0.2/certgen-linux-amd64 || echo "Certgen already in cache"
+          chmod +x ~/.minio/certs/certgen
+          cd ~/.minio/certs
+          ~/.minio/certs/certgen -ca -host "localhost"
       - name: Start a local instance of minio
         run: |
           export AWS_ACCESS_KEY_ID=Q3AM3UQ867SPQQA43P2F
@@ -49,11 +56,7 @@ jobs:
           export MINIO_ROOT_USER=Q3AM3UQ867SPQQA43P2F
           export MINIO_ROOT_PASSWORD=zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG
           chmod +x ~/cache/minio
-          chmod +x ~/cache/certgen
           mkdir -p ~/minio_tmp
-          mkdir -p ~/.minio/certs
-          cd ~/.minio/certs
-          ~/cache/certgen -ca -host "localhost,127.0.0.1,172.17.0.1"
           ~/cache/minio server ~/minio_tmp &
           sleep 4 # give minio some time to start
       - name: Run tests
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -705,14 +705,12 @@ class S3(object):
             headers['x-amz-server-side-encryption-aws-kms-key-id'] = self.config.kms_key
 
         if self.config.sse_customer_key:
-            md5s = md5()
-            sse_customer_key = self.config.sse_customer_key.encode()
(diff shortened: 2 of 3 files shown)


noonedeadpunk:
Sorry for spamming - I thought it would be easier to test out locally, but eventually some things that were passing locally were failing in CI.

Now I believe patch is in usable shape so you're welcome to review it!

S3/S3.py
```diff
@@ -8,28 +8,30 @@
 
 from __future__ import absolute_import, division
```
noonedeadpunk, line 10: I've just sorted imports - can revert them back

fviard:
> Sorry for spamming - I thought it would be easier to test out locally, but eventually some things that were passing locally were failing in CI.
> 
> Now I believe patch is in usable shape so you're welcome to review it!

Don't worry, it is always appreciated that you take the time to be sure that it is all good :-)

s3cmd
```diff
@@ -1920,6 +1923,15 @@ def cmd_sync_local2remote(args):
         error(u"or disable encryption with --no-encrypt parameter.")
         sys.exit(EX_USAGE)
 
+    # Disable md5 checks if using SSE-C. Add mtime check
+    if cfg.sse_customer_key or cfg.sse_copy_source_customer_key:
+        try:
+            cfg.sync_checks.remove("md5")
+        except Exception:
+            pass
+        if cfg.sync_checks.count("mtime") == 0:
```
fviard, line 1932: I think that you can simply do:
`if "mtime" not in cfg.sync_checks:`

S3/S3.py
```diff
@@ -701,6 +704,16 @@ def object_put(self, filename, uri, extra_headers = None, extra_label = ""):
             headers['x-amz-server-side-encryption'] = 'aws:kms'
             headers['x-amz-server-side-encryption-aws-kms-key-id'] = self.config.kms_key
 
+        if self.config.sse_customer_key:
+            md5s = md5()
+            sse_customer_key = self.config.sse_customer_key.encode()
+            md5s.update(sse_customer_key)
```
fviard, line 710: (Same for other similar blocks)
You can give the value to hash directly to the md5 like `md5(sse_customer_key)`.
So, I think that it will be cleaner if you pack everything in a single line and avoid temporary vars like:
`b64(md5(key).digest())`

S3/S3.py
```diff
@@ -701,6 +704,16 @@ def object_put(self, filename, uri, extra_headers = None, extra_label = ""):
...
+            md5_encoded = b64encode(md5s.digest())
+            key_encoded = b64encode(sse_customer_key)
+            headers["x-amz-server-side-encryption-customer-algorithm"] = "AES256"
+            headers["x-amz-server-side-encryption-customer-key"] = key_encoded.decode()
+            headers["x-amz-server-side-encryption-customer-key-md5"] = md5_encoded.decode()
```
fviard, line 715: I put the comment here, but same for all `decode()` and `encode()`:
Even if the effect will be the same, it would be better to use the `encode_to_s3()` and `decode_from_s3()` instead.

.github/workflows/test.yml
```diff
@@ -28,19 +28,32 @@ jobs:
...
-          export MINIO_ACCESS_KEY=Q3AM3UQ867SPQQA43P2F
-          export MINIO_SECRET_KEY=zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG
+          export MINIO_ROOT_USER=Q3AM3UQ867SPQQA43P2F
+          export MINIO_ROOT_PASSWORD=zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG
           chmod +x ~/cache/minio
+          chmod +x ~/cache/certgen
           mkdir -p ~/minio_tmp
+          mkdir -p ~/.minio/certs
+          cd ~/.minio/certs
+          ~/cache/certgen -ca -host "localhost,127.0.0.1,172.17.0.1"
```
fviard, line 56: Why do you add 172.17.0.1? localhost is not enough for github actions?

fviard:
I will need a little bit more time to review, but it looks very nice to me so far.

Your addition of "https" for the CI with certgen is a very good idea, i did not know about this tool.
Would it be possible that you extract this from this PR to put it in a separated one as it is a different subject?
Also, there is a little bit that surprise me, you don't setup the path to the certificates for Minio.
Does it use automatically certs generated with the same name from ~/.minio/certs forlder?

Also, to spare a few seconds of "cert" generation for nothing, maybe you can "cache" the generated certificates with certgen to avoid doing the expensive generation operation each time.

s3cmd
noonedeadpunk: Yeah, wasn't changing that part from fork. Changed

S3/S3.py
noonedeadpunk: Done

.github/workflows/test.yml
noonedeadpunk: Nah, it's not needed. it's just one of IPs minio listens on in CI, so just added all IPs while testing. Changed.

noonedeadpunk:
@fviard Um, yes, I can extract SSL from this PR. Another good thing to extract would be fixing of extra-headers that are currently not applied for object get.

But tbh I don't know how in github I can make series of patches. Ie, for this PR to pass CI, it should be based on top of the one with SSL. I dunno how to submit such series based on top of each other in github (super easy thing to do in gerrit :D)
So to make these separate PR and all of them working and passing CI sound to me close to impossible or I just lack knowledge:( Unless you can quick merge it and I then will be able to just rebase on master...

Regarding ~/.minio/certs - that is default path and is loaded automatically. Eventually certio is developed by minio as well, so they make it super easy to use: https://docs.min.io/docs/how-to-secure-access-to-minio-server-with-tls.html

noonedeadpunk:
Also te be fair, seems like my patch is kind of duplicating #1083

salanki:
Any movement on this one?

PR #1219, opened 2021-11-03
removed a default mutable argument pitfall

NaelsonDouglas:
**The problem**
In Python it usually dangerous to use mutable arguments like dicts or lists as default arguments in methods, as is better explained [here](https://docs.python-guide.org/writing/gotchas/)

**the solution**
This PR applies a simple refactoring to remove a case where a method was created using default a mutable argument


dtararukhin:
@NaelsonDouglas 
1) you missed `def test_s3cmd(label, cmd_args = [], **kwargs)`
2) there are no attempts to use lists modified by test(), so nothing is broken. Reviewers' policy looks quite conservative here, I bet on "no need" verdict :)

PR #1220, opened 2021-11-06
List support of Python 3.9 and 3.10

salty-horse:
Updated `setup.py` to list 3.9 and 3.10 (3.10 is passing the test suite so I assume it works fine)


fviard:
Merged, thanks!

Issue #212, opened 2013-08-12
Regression: --exclude '*.*' --rinclude '^one_file$' excludes all files


NV:
I want to upload several files, but don’t touch anything else:

```
s3cmd put --exclude '*.*' --rinclude '^one/file$' --rinclude '^another/file$'
```

Unfortunately, it doesn’t upload anything in 1.5.0-alpha3. It used to work well in 1.5.0-alpha1.

PR #1221, opened 2021-12-04
bugfix: --exclude dir/ walks inside dir/ for python 3.6+

dtararukhin:
I found out that `--ecxlude dir/` does not work properly on my system (Ubuntu 18.04 + python3.8) since it does not prevent from recursive scanning of all files inside excluded directory, which makes backuping big filesystems completely impossible. 

Here is why it happens:
see https://docs.python.org/3.5/library/fnmatch.html
```
>>> regex = fnmatch.translate('*.txt')
>>> regex
'.*\\.txt\\Z(?ms)'
```
compare with https://docs.python.org/3.6/library/fnmatch.html 
```
>>> regex = fnmatch.translate('*.txt')
>>> regex
'(?s:.*\\.txt)\\Z'
```
This change produces new regexp pattern for `--exclude dir/` which was not handled in FileLists.handle_exclude_include_walk_dir().

It's clear that program is broken and current behaviour is not expected:
1) function FileLists.handle_exclude_include_walk_dir doesn't do anything on python 3.6+
2) this comment https://github.com/s3tools/s3cmd/issues/212#issuecomment-32579310 says _"We absolutely need to be able to prune directories from the os.walk() via excludes."_

PR includes:
- fix the problem by adding new regexp pattern
- add test for that case

### DIFF
--- a/S3/FileLists.py
+++ b/S3/FileLists.py
@@ -60,15 +60,19 @@ def handle_exclude_include_walk_dir(root, dirname):
     Should this root/dirname directory be excluded? (otherwise included by default)
     Exclude dir matches in the current directory
     This prevents us from recursing down trees we know we want to ignore
-    return True for including, and False for excluding
+    return True for excluding, and False for including
     '''
     cfg = Config()
+    # python versions end their patterns (from globs) differently, test for different styles; check python3.6+ styles first
+    directory_patterns = (u'/)$', u'/)\\Z', u'\\/$', u'\\/\\Z(?ms)')
+
     d = os.path.join(root, dirname, '')
     debug(u"CHECK: %r" % d)
     excluded = False
     for r in cfg.exclude:
-        # python versions end their patterns (from globs) differently, test for both styles.
-        if not (r.pattern.endswith(u'\\/$') or r.pattern.endswith(u'\\/\\Z(?ms)')): continue # we only check for directory patterns here
+        if not any(r.pattern.endswith(dp) for dp in directory_patterns):
+            # we only check for directory patterns here
+            continue
         if r.search(d):
             excluded = True
             debug(u"EXCL-MATCH: '%s'" % (cfg.debug_exclude[r]))
@@ -76,8 +80,9 @@ def handle_exclude_include_walk_dir(root, dirname):
     if excluded:
         ## No need to check for --include if not excluded
         for r in cfg.include:
-            # python versions end their patterns (from globs) differently, test for both styles.
-            if not (r.pattern.endswith(u'\\/$') or r.pattern.endswith(u'\\/\\Z(?ms)')): continue # we only check for directory patterns here
+            if not any(r.pattern.endswith(dp) for dp in directory_patterns):
+                # we only check for directory patterns here
+                continue
             debug(u"INCL-TEST: %s ~ %s" % (d, r.pattern))
             if r.search(d):
                 excluded = False
--- a/run-tests.py
+++ b/run-tests.py
@@ -686,6 +686,14 @@ test_s3cmd("Sync remote2remote", ['sync', '%s/xyz/' % pbucket(1), '%s/copy/' % p
                   "delete: '%s/copy/etc/logo.png'" % pbucket(2) ],
     must_not_find = [ "blah.txt" ])
 
+## ====== Exclude directory
+test_s3cmd("Exclude directory", ['put', '-r', 'testsuite/demo/', pbucket(1) + '/xyz/demo/', '--exclude', 'dir1/', '-d'],
+           must_find = ["'testsuite/demo/dir2/file2-1.bin' -> '%s/xyz/demo/dir2/file2-1.bin'" % pbucket(1),
+                        "DEBUG: EXCLUDE: 'testsuite/demo/dir1/'"],  # whole directory is excluded
+           must_not_find = ["'testsuite/demo/dir1/file1-1.txt' -> '%s/xyz/demo/dir1/file1-1.txt'" % pbucket(1),
+                            "DEBUG: EXCLUDE: dir1/file1-1.txt"  # file is not synced, but also single file is not excluded
+                           ])
+
 ## ====== Don't Put symbolic link
 test_s3cmd("Don't put symbolic links", ['put', 'testsuite/etc/linked1.png', 's3://%s/xyz/' % bucket(1),],
            retcode = EX_USAGE,


dtararukhin:
Maybe I should also edit manpage, add better description of directory exclusion. It seems to me that logic of `--exclude dir/` clearly needs explanation with examples.

| #  | **command**                                           | **what it does**                                                                            | **os.walk inside excluded directory**                                                                                                                                                                                 |
|----|-------------------------------------------------------|---------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| 1  | `--exclude SomeDir`                                   | does not exclude anything in SomeDir/, only exclude file named SomeDir, if such file exists | Yes (if SomeDir is a directory)                                                                                                                                                                                       |
| 2  | `--exclude SomeDir/`                                  | exclude everything inside SomeDir/, all files and sub-directories                           | **No**. Exclude pattern ends with '/', so os.walk will stop at directory level.                                                                                                                                       |
| 3  | `--exclude SomeDir/*`                                 | exclude everything inside SomeDir/, all files and sub-directories                           | Yes. All files in SomeDir/ and sub-directories will be walked and excluded after one-by-one comparison with exclusion mask                                                                                            |
| 4  | `--rexclude SomeDir/`                                 | exclude everything inside SomeDir/, all files and sub-directories                           | Yes. All files in SomeDir/ and sub-directories will be walked and excluded after one-by-one comparison with exclusion mask                                                                                            |
| 5  | `--rexclude SomeDir/$`                                | does not exclude anything in SomeDir/                                                       | Yes. This regexp pattern matches only directory SomeDir/, but directory exclusion works only with `--exclude`, not with `--rexclude`                                                                                  |
| 6  | `--exclude SomeDir/ --include SomeDir/SomeFile`       | Exclude everything inside SomeDir/. SomeFile will also be excluded.                         | **No**. Exclude pattern ends with '/', so os.walk will stop at directory level. Include rule for SomeFile will not work                                                                                               |
| 7  | `--exclude SomeDir/* --include SomeDir/SomeFile`      | Exclude everything inside SomeDir/ except SomeFile                                          | Yes. All files in SomeDir/ and sub-directories will be walked and excluded after one-by-one comparison with exclusion mask. But SomeFile will be saved by `--include` option.                                         |
| 8  | `--rexclude SomeDir/ --include SomeDir/SomeFile`      | Exclude everything inside SomeDir/ except SomeFile                                          | Yes. All files in SomeDir/ and sub-directories will be walked and excluded after one-by-one comparison with exclusion mask. But SomeFile will be saved by `--include` option.                                         |
| 9  | `--exclude SomeDir/ --include SomeDir/SomeSubDir/`    | Exclude everything inside SomeDir/. SomeSubDir/ will also be excluded.                      | **No**. Exclude pattern ends with '/', so os.walk will stop at SomeDir/. Include rule for SomeDir/SomeSubDir/ will not work because program will not os.walk there. |
| 10 | `--rexclude SomeDir/ --include SomeDir/SomeSubDir/`   | Exclude everything inside SomeDir/. SomeSubDir/ will also be excluded.                      | Yes. All files in SomeDir/ and sub-directories will be walked and excluded after one-by-one comparison with exclusion mask. All files in SomeDir/SomeSubDir/ will not be saved since they don't match inclusion mask. |
| 11 | `--rexclude SomeDir/ --include SomeDir/SomeSubDir/*`  | Exclude everything inside SomeDir/ except SomeDir/SomeSubDir/ and its contents              | Yes. All files in SomeDir/ and sub-directories will be walked and excluded after one-by-one comparison with exclusion mask. But all files in SomeDir/SomeSubDir/ will be saved by inclusion mask.                     |
| 12 | `--rexclude SomeDir/ --rinclude SomeDir/SomeSubDir/`  | same as 11                                                                                  | same as 11                                                                                                                                                                                                            |

dtararukhin:
Hi @fviard, could you please take a look on this PR?
I want to merge this, update s3cmd package om my dev server and stop using program version from my virtualenv.

fviard:
Thank you very much for your very detailed contribution and sorry for the delay to review this.
You did a very good investigation, and it takes me a little bit of time to understand well.

The change in python behavior is very surprising, especially because it is not documented!

When I see the different regexp, globally, it is the "?m" that is different between the two of them.
But I don't understand how this could result in the difference related to the final "/" in the exclude string.
Do you have an idea?

fviard:
Your table is nice, it could go in the wiki maybe.
But, looking at that, it does not look normal to me that there is such a difference in behavior. for case like 1 and 2 for example.
Maybe we can improve that to be more deterministic.

S3/FileLists.py
```diff
@@ -60,34 +60,37 @@ def handle_exclude_include_walk_dir(root, dirname):
     Should this root/dirname directory be excluded? (otherwise included by default)
     Exclude dir matches in the current directory
     This prevents us from recursing down trees we know we want to ignore
-    return True for including, and False for excluding
+    return True for excluding, and False for including
     '''
     cfg = Config()
+    # python versions end their patterns (from globs) differently, test for different styles
+    directory_patterns = [u'\\/$', u'\\/\\Z(?ms)', u'/)$', u'/)\\Z']
```
fviard, line 67: it is a micro optimisation, but you can put the new pattern before in the list as it should be the one that will match most with recent versions.

S3/FileLists.py
```diff
@@ -60,34 +60,37 @@ def handle_exclude_include_walk_dir(root, dirname):
...
+
     d = os.path.join(root, dirname, '')
     debug(u"CHECK: %r" % d)
     excluded = False
     for r in cfg.exclude:
-        # python versions end their patterns (from globs) differently, test for both styles.
-        if not (r.pattern.endswith(u'\\/$') or r.pattern.endswith(u'\\/\\Z(?ms)')): continue # we only check for directory patterns here
+        if not any(r.pattern.endswith(dp) for dp in directory_patterns): continue # we only check for directory patterns here
```
fviard, line 73: You can profit to clean up a bit the code by putting the continue and the comment on a new line :-)

S3/FileLists.py
```diff
@@ -60,34 +60,37 @@ def handle_exclude_include_walk_dir(root, dirname):
...
+        debug(u"EXCL-TEST: %s ~ %s" % (d, r.pattern))
```
fviard, line 74: Maybe this new debug is not necessary once it works .

S3/FileLists.py
```diff
@@ -60,34 +60,37 @@ def handle_exclude_include_walk_dir(root, dirname):
...
             debug(u"INCL-TEST: %s ~ %s" % (d, r.pattern))
             if r.search(d):
                 excluded = False
                 debug(u"INCL-MATCH: '%s'" % (cfg.debug_include[r]))
                 break
     if excluded:
         ## Still excluded - ok, action it
-        debug(u"EXCLUDE: %r" % d)
+        ## Use relative name to easily check this in tests
+        debug(u"EXCLUDE: %r" % dirname)
```
fviard, line 91: Interesting idea, but still I think that it is better to have the full name.
Because if debugging the behavior of s3cmd, it might be easier to understand an issue if we know where is the folder that is excluded.

S3/FileLists.py
dtararukhin: 1) Here I just copied same logic for debug messages as in current code, see INCL-TEST/INCL-MATCH messages below on lines 81 and 84 (old numeration). 
2) in general, all debugs are unnecessary when everything works fine :) This one was really helpful when I was trying to figure out what happens: does program escape from cycle on `if not any(r.pattern.endswith... continue` one line above, or we do check regexps, but they just don't match.

S3/FileLists.py
dtararukhin: > it is better to have the full name

Yes, it's reasonable. 

But then I'll need to fix my "Exclude directory" test, because EXCLUDE debug messages will contain absolute paths that depend on local environment. I can use regexps (must_find_re/must_not_find_re) with broad patterns (`.*`) instead of simple string comparison (must_find/must_not_find), this way doesn't look pretty for me, but if there's no other way, then I'll do it. What do you think?

dtararukhin:
> But, looking at that, it does not look normal to me that there is such a difference in behavior. for case like 1 and 2 for example. 
Maybe we can improve that to be more deterministic.

I agree, current behaviour is really tricky. I was thinking a lot about it, at certain point I was even sure that all include/exclude  part should be completely refactored :) But s3cmd tool is quite popular, and after some time I decided that the best strategy for my PR is fixing only things that are really broken, without touching anything else, to avoid any unexpected effects for people who currently use these options. So, I'd suggest straightening include/exclude logic in separate PR.

S3/FileLists.py
fviard: It is better to have the code more helpful and the test less good than the opposite.

For the test, the following regexp should be enough, no?
'testsuite/demo/dir1$'
otherwise .* is ok without issue.

S3/FileLists.py
fviard: I don't like very much the log here because, in the other cases, we have already an "exclusion / inclusion Match" before printing a log.


The EXCL-TEST would print the line for all test for all directories tested.
In my opinion it is only useful to debug the current issue, and could be temporary added like now to debug if necessary.

And, in the end, you will have the PASS and EXCLUDE log that will give you an hint if something does not go as expected.

dtararukhin: ok, removed

S3/FileLists.py
dtararukhin: ok, got it
also, 1) I was wrong about full path in debug messages, actually I don't need any regexp in my test, 2) found and fixed a bug in my test (wrong file names in must_not_find list)

Now everything looks fine

dtararukhin:
@fviard please check: I think I've made all requested changes

S3/FileLists.py
```diff
@@ -60,24 +60,29 @@ def handle_exclude_include_walk_dir(root, dirname):
     Should this root/dirname directory be excluded? (otherwise included by default)
     Exclude dir matches in the current directory
     This prevents us from recursing down trees we know we want to ignore
-    return True for including, and False for excluding
+    return True for excluding, and False for including
     '''
     cfg = Config()
+    # python versions end their patterns (from globs) differently, test for different styles; check python3.6+ styles first
+    directory_patterns = [u'/)$', u'/)\\Z', u'\\/$', u'\\/\\Z(?ms)']
```
fviard, line 67: A last small thing, could you transform that into a "tuple" instead of a list as it will not be dynamically changed?

fviard:
Also, in the end, could you squash your commits into a single commit? and then it will be all good to be merged :-) thanks!

dtararukhin:
@fviard Everything is done!

fviard:
Merged, thank you very much!

dtararukhin:
@fviard cool, thanks!
Also, do you have a plan to make new release? I tried to reinstall s3cmd on my dev server using `pip install` but obviously I get 2.2.0 version as of 2021-09-27 from pypi.org.

PR #1227, opened 2022-01-14
Take into account the AWS_STS_REGIONAL_ENDPOINTS

lanzrein:
This fixes the issue where the sts endpoint sts.amazonaws.com is not
accessible.
Take into account the environment varialbe AWS_STS_REGIONAL_ENDPOINTS
If the AWS_STS_REGIONAL_ENDPOINTS variable is provided check for AWS_REGION env var to decide
region.

If AWS_REGION is not provided use the same region as the bucket.
(bucket_location parameter)

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -318,7 +318,15 @@ class Config(object):
                     '%s=%s' % (k, s3_quote(v, unicode_output=True))
                     for k, v in params.items()
                 ])
-                conn = httplib.HTTPSConnection(host='sts.amazonaws.com',
+                sts_endpoint = "sts.amazonaws.com"
+                if os.environ.get("AWS_STS_REGIONAL_ENDPOINTS") == "regional": 
+                    # Check if the AWS_REGION variable is available to use as a region. 
+                    region = os.environ.get("AWS_REGION")
+                    if not region: 
+                        # Otherwise use the bucket location
+                        region = self.bucket_location
+                    sts_endpoint = "sts.%s.amazonaws.com" % region 
+                conn = httplib.HTTPSConnection(host=sts_endpoint,
                                                timeout=2)
                 conn.request('POST', '/?' + encoded_params)
                 resp = conn.getresponse()


aperigault:
Hi,
tested here and it solved my problem.

Thanks !

S3/Config.py
```diff
@@ -318,7 +318,16 @@ def role_config(self):
...
-                conn = httplib.HTTPSConnection(host='sts.amazonaws.com',
+                sts_endpoint = "sts.amazonaws.com"
+                if os.environ.get("AWS_STS_REGIONAL_ENDPOINTS") == "regional": 
+                    # Check if the AWS_REGION variable is available to use as a region. 
+                    if os.environ.get("AWS_REGION"): 
+                        region = os.environ.get("AWS_REGION")
+                    else : 
+                        # Otherwise use the bucket location
+                        region = self.bucket_location
+                    sts_endpoint = f"sts.{region}.amazonaws.com"
```
fviard, line 329: Can you use %-based string instead of f-string for retro compatibility?

S3/Config.py
```diff
@@ -318,7 +318,16 @@ def role_config(self):
                     '%s=%s' % (k, s3_quote(v, unicode_output=True))
                     for k, v in params.items()
                 ])
-                conn = httplib.HTTPSConnection(host='sts.amazonaws.com',
+                sts_endpoint = "sts.amazonaws.com"
+                if os.environ.get("AWS_STS_REGIONAL_ENDPOINTS") == "regional": 
+                    # Check if the AWS_REGION variable is available to use as a region. 
+                    if os.environ.get("AWS_REGION"): 
+                        region = os.environ.get("AWS_REGION")
```
fviard, line 325: Instead of getting the value 2 times, you can do:
```
region = os.env...
if not region:
    region = self.bucket_location
```

S3/Config.py
lanzrein: Hello, this has been fixed. 
I used the following : 
```python 
sts_endpoint = "sts.%s.amazonaws.com" % region
```

Let me know if this is how you wanted it.

S3/Config.py
lanzrein: Done thanks for the feedback !

fviard:
It's all good now, thank you.

Could you "squash" your changes into a single commit before I merge please?

lanzrein:
Should be all good !

fviard:
Merged, thank you!

aperigault:
Thank you @fviard. Can we have a release with this fix included please ?

PR #1232, opened 2022-01-31
cmd_sync_remote2local handle 503 Slow Down

Kaffeetasse:
Hello, my cloud provider also limits downloads, so I needed this feature.


fviard:
Thank you very much for your contribution.

I'm closing this MR because I just pushed a change to handle this a little bit cleaner for all download commands:
https://github.com/s3tools/s3cmd/commit/0922f1e1ff626cfd125d12053918363901d671d5

This change should fix the issue with a "retry after a delay", but it will not "throttle" all the subsequent requests.
I'm not sure if it was what was needed for your case, but if ever we would need to "throttle" the delay between requests, it is probably a config in the same way as limitrate that would be needed. But having to use that would be very disappointing for a S3 service.

Issue #1005, opened 2018-09-26
ValueError: error:060800A3:digital envelope routines:EVP_DigestInit_ex:disabled for fips


kamal2222ahmed:
I am trying to use s3cmd on CentOs 7.5 , and it works fine with $s3cmd ls s3://alpha-team-share/chefprod

but when uploading:

s3cmd put backup_chef_2018-09-26-133542.tar.gz s3://alpha-team-share/chefprod

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
    An unexpected error has occurred.
  Please try reproducing the error using
  the latest s3cmd code from the git master
  branch found at:
    https://github.com/s3tools/s3cmd
  and have a look at the known issues list:
    https://github.com/s3tools/s3cmd/wiki/Common-known-issues-and-their-solutions
  If the error persists, please report the
  following lines (removing any private
  info as necessary) to:
   s3tools-bugs@lists.sourceforge.net

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

Invoked as: /bin/s3cmd put backup_chef_2018-09-26-133542.tar.gz s3://alpha-team-share/chefprod
Problem: ValueError: error:060800A3:digital envelope routines:EVP_DigestInit_ex:disabled for fips
S3cmd:   2.0.2
python:   2.7.5 (default, May 31 2018, 09:41:32)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-28)]
environment LANG=en_US.UTF-8

Traceback (most recent call last):
  File "/bin/s3cmd", line 3092, in <module>
    rc = main()
  File "/bin/s3cmd", line 3001, in main
    rc = cmd_func(args)
  File "/bin/s3cmd", line 369, in cmd_object_put
    local_list, single_file_local, exclude_list, total_size_local = fetch_local_list(args, is_src = True)
  File "/usr/lib/python2.7/site-packages/S3/FileLists.py", line 352, in fetch_local_list
    total_size = _fetch_local_list_info(local_list)
  File "/usr/lib/python2.7/site-packages/S3/FileLists.py", line 231, in _fetch_local_list_info
    md5 = loc_list.get_md5(relative_file) # this does the file I/O
  File "/usr/lib/python2.7/site-packages/S3/FileDict.py", line 48, in get_md5
    md5 = Utils.hash_file_md5(self[relative_file]['full_name'])
[...]

PR #1233, opened 2022-02-02
Add FIPS support

maroth96:
When FIPS mode is enabled, s3cmd raises a ValueError exception with the message 'digital envelope routines: EVP_DigestInit_ex] disabled for FIPS' (see below). This change will make s3cmd compatible with FIPS mode.

```
$ fips-mode-setup --check
FIPS mode is enabled.

$ s3cmd put ...
  ...
Problem: <class 'ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS
S3cmd:   2.2.0
python:   3.6.8 (default, Jan 19 2022, 23:28:49) 
[GCC 8.5.0 20210514 (Red Hat 8.5.0-7)]
environment LANG=en_US.UTF-8
  File "./s3cmd/s3cmd", line 3209, in <module>
    rc = main()
  File "./s3cmd/s3cmd", line 3106, in main
    rc = cmd_func(args)
  File "./s3cmd/s3cmd", line 392, in cmd_object_put
    local_list, single_file_local, exclude_list, total_size_local = fetch_local_list(args, is_src = True)
  File "/home/michaelroth/s3cmd/S3/FileLists.py", line 367, in fetch_local_list
    total_size = _fetch_local_list_info(local_list)
  File "/home/michaelroth/s3cmd/S3/FileLists.py", line 238, in _fetch_local_list_info
    md5 = loc_list.get_md5(relative_file) # this does the file I/O
  File "/home/michaelroth/s3cmd/S3/FileDict.py", line 48, in get_md5
    md5 = Utils.hash_file_md5(self[relative_file]['full_name'])
  File "/home/michaelroth/s3cmd/S3/Utils.py", line 105, in hash_file_md5
    h = md5()
ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS
```

### DIFF
--- a/S3/BaseUtils.py
+++ b/S3/BaseUtils.py
@@ -8,6 +8,8 @@
 
 from __future__ import absolute_import, division
 
+import functools
+import hashlib
 import re
 import sys
 
@@ -48,6 +50,19 @@ except NameError:
     # python 3 support
     # In python 3, unicode -> str, and str -> bytes
     unicode = str
+
+
+try:
+   md5 = hashlib.md5()
+except Exception as exc:
+  try:
+    # md5 is disabled for FIPS-compliant Python builds.
+    # Since s3cmd does not use md5 in a security context,
+    # it is safe to allow the use of it by setting useforsecurity to False.
+    hashlib.md5(usedforsecurity=False)
+    md5 = functools.partial(hashlib.md5, usedforsecurity=False)
+  except:
+      raise exc
 
 
 __all__ = []
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -33,13 +33,8 @@ except ImportError:
 
 import select
 
-try:
-    from hashlib import md5
-except ImportError:
-    from md5 import md5
-
 from .BaseUtils import (getListFromXml, getTextFromXml, getRootTagName,
-                        decode_from_s3, encode_to_s3, s3_quote)
+                        decode_from_s3, encode_to_s3, md5, s3_quote)
 from .Utils import (convertHeaderTupleListToDict, hash_file_md5, unicodise,
                     deunicodise, check_bucket_name,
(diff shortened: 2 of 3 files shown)


maroth96:
This should fix this issue: https://github.com/s3tools/s3cmd/issues/1005

fviard:
Thank you for your PR, I never heard about this argument for hashlib but it is awesome to unblock people in such a case.

I notice here that it is only available since python 3.9 or on Fedora/redhat distributions. But still better than nothing.

I would suggest the following improvement to your PR:
As the md5 will be used a lot of times and also in multiple places in the code, you could do something like:

- Do the try/except just once inside BaseUtils.py. Something like:
```
from functools import partial
from hashlib import md5_real
try:
   md5 = md5_real()
except Exception as exc:
  # Comment to explain that it is for fips
  try:
    md5_real(usedforsecurity=False)
    md5 = partial(md5_real, usedforsecurity=False)
  except:
      # Raise original issue
      raise exc
```

- import md5 from BaseUtils.py and use it directly in the other places of s3cmd

maroth96:
Done. md5 is moved to BaseUtils.

maroth96:
@fviard Is this change OK for submission.

fviard:
Merged, thank you very much and sorry for the delay. I think that I did not notice that you pushed the requested changes.

Btw, I will modify further the code in another commit, because as the "usedforsecurity" will be available upstream, it will be good to have that as a default case.

uipatherthangaraje:
When is the plan to release it? how can I consume this fix? thanks

rajivml:
@fviard glad this is fixed but I couldn't find a release post october 3rd where this fix is available  https://github.com/s3tools/s3cmd/releases

PR #1245, opened 2022-03-25
Add setobjectretention and setobjectlegalhold commands

withings-sas:
These commands enable s3cmd to activate the 'retention' and 'legal hold' on an object.

I tried to keep it very simple and flexible, so the xml file containing the arguments has to be written outside and given as an argument.

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -1063,6 +1063,26 @@ class S3(object):
         response = self.send_request(request)
         return decode_from_s3(response['data'])
 
+    def set_object_legal_hold(self, uri, legal_hold):
+        headers = SortedDict(ignore_case = True)
+        headers['content-type'] = 'application/xml'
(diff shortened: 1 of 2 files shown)


withings-sas:
Any chance someone can review this ? It's quite straighforward but please let us know if we need to make some improvements

fviard:
Thank you for your contribution and sorry to give you a feedback so late.
These are interesting new commands to add to s3cmd.

Regarding the fact that they take only a very limited input, (a single "ON/OFF" parameter for setobjectlogalhold for example), would you be ok to rework your patch to just do the requested action directly, without taking a xml file as input?

In my mind, it is different of a policy for example, that could be quite complex.

Also, it would be interesting to have the "get" command going with the "set" I think :-)

fviard:
Closing as replaced by #1297 .

PR #1255, opened 2022-05-18
Use IMDSv2 to retrieve instance metadata

AnthonyFoiani-at:
# Summary

Hello!

We are trying to disable IMDSv1 across our fleet, to improve our security posture.

(Some background: an explanation of what AWS did, and why, is in [Add defense in depth against open firewalls, reverse proxies, and SSRF vulnerabilities with enhancements to the EC2 Instance Metadata Service](https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/). They also provide a helpful article on how to [Use IMDSv2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html).)

`s3cmd` had IMDSv1 hard-coded, so it failed on machines where we disabled this feature.

This PR is a (fairly minimal?) change to use IMDSv2. (It would probably be much cleaner to use the AWS-supported APIs, but I also appreciate that `s3cmd` is a nice standalone install!)

Some other folks have hit this problem:

* Gitlab Chart: [Cloud native backup does not support deployments in AWS with IMDSv1 disabled](https://gitlab.com/gitlab-org/charts/gitlab/-/issues/2787)

# Testing

With original code, on a machine with IMDSv1 disabled:

```
root@ip-172-20-137-224:/usr/local/lib/python2.7/dist-packages/S3# ln -s Config-orig.py Config.py
root@ip-172-20-137-224:/usr/local/lib/python2.7/dist-packages/S3# s3cmd help
ERROR: /root/.s3cfg: None
ERROR: Configuration file not available.
ERROR: Consider using --configure parameter to create one.
```

With this fix:

```
root@ip-172-20-137-224:/usr/local/lib/python2.7/dist-packages/S3# ln -s Config-imdsv2.py Config.py
root@ip-172-20-137-224:/usr/local/lib/python2.7/dist-packages/S3# s3cmd help
ERROR: Invalid command: help
```

Please consider rolling this in to your tool. Thanks!

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -354,11 +354,30 @@ class Config(object):
             else:
                 conn = httplib.HTTPConnection(host='169.254.169.254',
                                               timeout=2)
-                conn.request('GET', "/latest/meta-data/iam/security-credentials/")
+
+                # To use Instance Metadata Service (IMDSv2), we first need to obtain a token, then
+                # supply it with every IMDS HTTP call. More info:
+                #
+                #   https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html
+                #
+                # 60 seconds is arbitrary, but since we're just pulling small bits of data from the
+                # local instance, it should be plenty of time.
+                #
+                imds_ttl = {"X-aws-ec2-metadata-token-ttl-seconds": "60"}
+                conn.request('PUT', "/latest/api/token", headers=imds_ttl)
+                resp = conn.getresponse()
+                if resp.status != 200:
+                    raise IOError
+                imds_token = resp.read().decode('utf-8')
+                imds_auth = {"X-aws-ec2-metadata-token": imds_token}
+
+                conn.request('GET', "/latest/meta-data/iam/security-credentials/", headers=imds_auth)
                 resp = conn.getresponse()
                 files = resp.read()
                 if resp.status == 200 and len(files) > 1:
-                    conn.request('GET', "/latest/meta-data/iam/security-credentials/%s" % files.decode('utf-8'))
+                    conn.request('GET',
+                                 "/latest/meta-data/iam/security-credentials/%s" % files.decode('utf-8'),
+                                 headers=imds_auth)
                     resp=conn.getresponse()
                     if resp.status == 200:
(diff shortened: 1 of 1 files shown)


fviard:
Thank you for the info and the PR.

"The AWS approach is to block open WAFs by using a type of request that open WAFs very rarely support, HTTP PUT requests."
I'm not really convinced by the added value of IMDSv2 compare to the burden, but indeed it is very good to add support for that to s3cmd.

Your change looks good also, but there is one remaining question to resolve before merging:
- Is there any s3 compatible service that would also offer IMDSv1 currently, doing the same thing as amazon, but that would not support IMDSv2 requests?
- If yes, how can we avoid breaking them and remain compatible? Maybe we should just ignore a failure to /latest/api/token, and only set "X-aws-ec2-metadata-token" if we successfully got one?

AnthonyFoiani-at:
> Thank you for the info and the PR.

You're very welcome! Thank **you** for maintaining a great tool!

> Is there any s3 compatible service that would also offer IMDSv1 currently, doing the same thing as amazon, but that would not support IMDSv2 requests?

I think it's more than just s3 compatibility that would have to be present; it would be the deep structure of instance metadata and credentials.

I could see an S3 compatibility issue where the various credentials used by some otherwise-compatible-with-s3 service need to get set -- but I would expect the intersection of "emulates `ec2metadata` v1 correctly" and "doesn't support `ec2metadata` v2 properly" is pretty small. :-)

Regardless, your suggestion sounds good to me. Let me bake that in; worst case, we don't get a token, so we try without the token.

Hopefully can update the PR in a few minutes!

AnthonyFoiani-at:
> Maybe we should just ignore a failure to /latest/api/token, and only set "X-aws-ec2-metadata-token" if we successfully got one?

Hopefully https://github.com/s3tools/s3cmd/pull/1255/commits/4b7ad6de7d276e4f380d7de25ca9a1fa8e7f04ff is close to what you had in mind?

A quick test with the same technique succeeded in the same way.

fviard:
Merged, thank you!

AnthonyFoiani-at:
Thank **you**!

PR #1258, opened 2022-05-30
feat: Added bi cmd

ethiebaut:
Added a new feature to download a CSV file of the bucket content, useful for local analysis


fviard:
Thank you for your contribution, but I'm not planing to merge this change, because it is a little bit too specific to one use case.

Normally, you can easily parse the "text" output to generate a csv. Or, otherwise, use s3cmd as a library (ie "S3") to do that.

It might not be bad to have "machine parsable" output in s3cmd, but then it should be a general rework that will cover most "get" commands in an uniform way.

Issue #1269, opened 2022-07-23
[Feature Request]: list-objects-v2 allow unordered support.


snosratiershad:
In ceph storage's s3 interface, there is a flag named "allow-unordered" that allows rgw (the gateway) to don't force the cluster to sort all of objects in case of list-objects-v2 requests.
There is also a command option in radosgw-admin (gateway commandline utility) named "--allow-unordered" which provides unordered list to.
It's not officially supported by AWS, but included in S3 services which are based on Ceph storage (for example, DigitalOcean Spaces, Redhat Ceph Storage, Alibaba Cloud OSS, Sotoon S3, ...)

Is there any plan to support unordered list in s3cmd?

Update: I found a way to extend aws-cli's with add [this file](https://github.com/ceph/ceph/blob/main/examples/boto3/service-2.sdk-extras.json) to `~/.aws/models/s3/2006-03-01/`. but not available in s3cmd.

fviard:
@snosratiershad Are you sure that it is supported for the list-objets also?

In ceph documentation, I only see this option available in the BucketOps doc page and not in ObjectOps page.

Where you able to test this parameter with DigitalOcean?

snosratiershad:
@fviard, yes! it works on list-objects and list-objects-v2 (and I recently added list-objects-v2 support to boto3 (and aws) ceph extention [PR](https://github.com/ceph/ceph/pull/47487))
I will search for ceph documentation, but I pretty sure about that, and I use it every day, maintaining 3 ceph clusters with PBs of objects.
about digitalocean I've not tested it by my own, but as it's based on ceph, I guess it should handle it too.

fviard:
Ok, good news.  
I can look at adding that, does not cost much I think.

So, to be sure to understand well, you just have to add "allow-unordered" to query param of the list-objects* to have this behavior?

Btw, it has been a long time that I did not have a look at aws-cli, but I would not recommend you to use that option with their sync, as they are relying on the fact that keys are order to decide on which file to synchronize.

snosratiershad:
@fviard 
> So, to be sure to understand well, you just have to add "allow-unordered" to query param of the list-objects* to have this behavior?

yes!


> Btw, it has been a long time that I did not have a look at aws-cli, but I would not recommend you to use that option with their sync, as they are relying on the fact that keys are order to decide on which file to synchronize. 

[...]

PR #1270, opened 2022-08-08
list-objects-*: implement allow unordered option

snosratiershad:
As discussed in [this issue](https://github.com/s3tools/s3cmd/issues/1269)

### DIFF
--- a/.ci.s3cfg
+++ b/.ci.s3cfg
@@ -39,6 +39,7 @@ kms_key =
 limit = -1
 limitrate = 0
 list_md5 = False
+list_allow_unordered = False
 log_target_prefix = 
 long_listing = False
 max_delete = -1
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -247,7 +247,7 @@ class Config(object):
     # Not an AWS standard
     # allow the listing results to be returned in unsorted order.
     # This may be faster when listing very large buckets.
-    list_objects_allow_unordered = False
+    list_allow_unordered = False
 
     ## Creating a singleton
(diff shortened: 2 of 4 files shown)


snosratiershad:
@fviard, if there is anything I can help about it, feel free to contact me.

fviard:
Sorry for the delay of my reply.  
Instead of adding the parameter to all the function prototypes, i think that you can use directly `self.config.allow_unordered` inside ` def bucket_list_noparse(...)`.

snosratiershad:
@fviard, as you requested, I changed it to a config variable.
I couldn't found any of documentation files in the repository, but I think it must be included in somewhere as there is no help message for now.

snosratiershad:
@fviard, should command line switch literally overwrite config file, or just ignoring the value comes from `self.config.list_objects_allow_unordered` when the `--ls-allow-unordered'  switch called?

snosratiershad:
I realized what you mean. also going to change "list-objects-allow-unordered" to "list-allow-onordered".

snosratiershad:
@fviard Done.

fviard:
Merged, thank you!

visbits:
Delimiter cannot be used with allow-unordered.

The code must not set a limiter if allow-unordered is true:
From line 400 in S3.py
`        if not self.config.recursive and not recursive:`
To
`        if not self.config.recursive and not recursive and not self.config.list_allow_unordered:`

snosratiershad:
@visbits Thanks for pointing this out. I'll check it out and fix it.

Issue #1271, opened 2022-08-08
Feature request: ListObjectsV2 request option in `ls` command


snosratiershad:
In progress of [this PR](https://github.com/s3tools/s3cmd/pull/1270) I noticed s3cmd doesn't implemented the ListObjectsV2 requests and all operations in `ls` command is under ListObjects.

Is there any plan to support it? let me know and I would create a PR ASAP.

snosratiershad:
@fviard, do you think it's also should be included as a config variable like `list_version`? or a command prompt param like `--list-objects-v2` as aws cli?

fviard:
I don't think that there is a big interest in switching to ListObjectV2 for performance or anything, and it is probably not possible to switch as S3 compatible servers might not have migrated to that yet.

But, it could be interesting to already have support for ListObjectV2 to be future proof and for a possible server that would not support ListObjectV1.

So, it would be good if you want to add the support for that in s3cmd :-)  
I don't think that a cmdline parameter is useful, but we can probably have a config variable. Something like "use_list_object_v2" or "enable_list_object_v2".

snosratiershad:
on it!

snosratiershad:
@fviard Done.

showvick:
Amazon recently added Amazon S3 Express One Zone. This is a high-performance, single-Availability Zone storage class purpose-built to deliver consistent single-digit millisecond data access for your most frequently accessed data and latency-sensitive applications. S3 Express One Zone can improve data access speeds by 10x and reduce request costs by 50% compared to S3 Standard.

Amazon S3 Express One Zone doesn't support ListObjects. It supports ListObjectsV2 API. Hence S3cmd command cannot be used with S3 express one zone directory. It fails as below. It would be good if this functionality can be implemented. 

# s3cmd ls -r -p s3://<S3 Express One Zone bucket>
ERROR: S3 error: 501 (NotSupported): This bucket does not support ListObjects API. Consider using ListObjectsV2 API.

snosratiershad:
I just submitted a PR on this issue

PR #1274, opened 2022-08-28
feat: implement ListObjectsV2 support

snosratiershad:
as discussed in [this issue](https://github.com/s3tools/s3cmd/issues/1271).

### DIFF
--- a/.ci.s3cfg
+++ b/.ci.s3cfg
@@ -15,6 +15,7 @@ delete_after = False
 delete_after_fetch = False
 delete_removed = False
 dry_run = False
+enable_list_objects_v2 = False
 enable_multipart = True
 encoding = UTF-8
 encrypt = False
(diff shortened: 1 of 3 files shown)


fviard:
Thanks for the PR.
But I don't think that it is enough to just use the version 2 as-is. I think a few fields changed also. Like 'marker' is not available anymore, but replaced by "start after" that is somewhat equivalent?

There is now also the "continuation token" that it might be more interesting for us to use than the "marker" of the v1.

fviard:
I'm closing this MR for the moment as it is incomplete.
But don't hesitate to reopen if ever you would have a more complete solution for this in the future.

snosratiershad:
@fviard, sorry for my not being responsible for this PR. I'm going to reopen that in a new PR this week.

PR #1276, opened 2022-09-10
Improve error messages when S3 returns a 403 error during an `ls` operation

qris:
Include the bucket name and URI in the logged error message (via the exception object).

Before:

    ERROR: S3 error: 403 (Forbidden)

After:

    ERROR: S3 error: 403 (Forbidden): {'bucket': 'iPhoto....net', 'uri': '/.../Masters/2016/10/15/20161015-142339/IMG_3290.MOV'}

### DIFF
--- a/S3/Exceptions.py
+++ b/S3/Exceptions.py
@@ -136,6 +136,20 @@ class S3Error (S3Exception):
             raise S3ResponseError("Malformed error XML returned from remote server.")
         return info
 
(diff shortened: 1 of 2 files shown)


fviard:
Have you an example of command and debug output ("-d") that will generate a case like that?
Normally, you should already have access to the info through the existing exc.resource, no? Except if aws does not return it, but, then it maybe means that the object in itself is not what is causing the issue.

PR #1278, opened 2022-10-04
fix: Use NextMarker if only prefixes are returned

Fumesover:
Currently, if s3cmd is used to list objects in bucket, and if only CommonPrefixes are returned, the cli will ignore the NextMarker field and will use the last given prefix.

This fix will consider using the NextMarker if it's present, and will fallback on the last given prefix in both cases. Copying the workflow of awscli


fviard:
Merged, thank you very much!

PR #1279, opened 2022-10-06
Re-work bucket naming rules.

blkmajik:
Add bucket_name_quirks config option (boolean).  This allows you to use a colon in the bucket name for Ceph compatibility with tenants.

Since it's after March 1, 2018 apply the bucket naming rules for bucket creation across the board.  Still allow dns_strict to be used if we ever check bucket names for not creation events.  See

   https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html

for details.

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -133,6 +133,8 @@ class Config(object):
     force = False
     server_side_encryption = False
     enable = None
+    # Used to allow colons in bucket names for Ceph compatibility
+    bucket_name_quirks = False
     get_continue = False
     put_continue = False
     upload_id = u""
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -407,17 +407,21 @@ class S3(object):
             headers.update(extra_headers)
 
         body = ""
-        if bucket_location and bucket_location.strip().upper() != "US" and bucket_location.strip().lower() != "us-east-1":
-            bucket_location = bucket_location.strip()
-            if bucket_location.upper() == "EU":
-                bucket_location = bucket_location.upper()
-            body  = "<CreateBucketConfiguration><LocationConstraint>"
-            body += bucket_location
-            body += "</LocationConstraint></CreateBucketConfiguration>"
-            debug("bucket_location: " + body)
-            check_bucket_name(bucket, dns_strict = True)
+        if self.config.bucket_name_quirks:
+            # We are explicitly not AWS
+            check_bucket_name(bucket, dns_strict = False, name_quirks = True)
         else:
-            check_bucket_name(bucket, dns_strict = False)
+            if bucket_location:
+                # We follow AWS rules
+                bucket_location = bucket_location.strip()
+                if bucket_location.upper() == "EU":
+                    bucket_location = bucket_location.upper()
+                body  = "<CreateBucketConfiguration><LocationConstraint>"
+                body += bucket_location
+                body += "</LocationConstraint></CreateBucketConfiguration>"
+                debug("bucket_location: " + body)
+            check_bucket_name(bucket, dns_strict = True, name_quirks = False)
+
         if self.config.acl_public:
             headers["x-amz-acl"] = "public-read"
 
--- a/S3/Utils.py
+++ b/S3/Utils.py
@@ -234,8 +234,30 @@ def time_to_epoch(t):
     raise S3.Exceptions.ParameterError('Unable to convert %r to an epoch time. Pass an epoch time. Try `date -d \'now + 1 year\' +%%s` (shell) or time.mktime (Python).' % t)
 
 
-def check_bucket_name(bucket, dns_strict=True):
-    if dns_strict:
+def check_bucket_name(bucket, dns_strict=True, name_quirks=False):
+    """Check that the bucket name is valid for our situation.
+
+    Check against the rules from: https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html
+
+    dns_strict: True means follow above rules exactly.   False allows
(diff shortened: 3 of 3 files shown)


S3/Utils.py
```diff
@@ -244,27 +266,29 @@ def check_bucket_name(bucket, dns_strict=True):
...
-            raise S3.Exceptions.ParameterError("Bucket name '%s' must not contain sequence '-.' for DNS compatibility" % bucket)
-        if re.search("\.\.", bucket, re.UNICODE):
-            raise S3.Exceptions.ParameterError("Bucket name '%s' must not contain sequence '..' for DNS compatibility" % bucket)
-        if not re.search("^[0-9a-z]", bucket, re.UNICODE):
-            raise S3.Exceptions.ParameterError("Bucket name '%s' must start with a letter or a digit" % bucket)
-        if not re.search("[0-9a-z]$", bucket, re.UNICODE):
-            raise S3.Exceptions.ParameterError("Bucket name '%s' must end with a letter or a digit" % bucket)
+    if len(bucket) > max_length:
+        raise S3.Exceptions.ParameterError("Bucket name '%s' is too long (max %d characters)" % (bucket, max_length))
+    if re.search("-\.", bucket, re.UNICODE):
```
fviard, line 277: You can profit of the rework to simplify this to:
`if '-.' in bucket`

S3/Utils.py
```diff
@@ -244,27 +266,29 @@ def check_bucket_name(bucket, dns_strict=True):
...
+        raise S3.Exceptions.ParameterError("Bucket name '%s' must not contain sequence '-.' for DNS compatibility" % bucket)
+    if re.search("\.\.", bucket, re.UNICODE):
```
fviard, line 279: Same a previous one, you could probably simplify that in:
`if '..' in bucket`

And maybe merge the 2 tests in fact!

S3/Utils.py
```diff
@@ -234,8 +234,30 @@ def time_to_epoch(t):
...
+    if name_quirks:
+        invalid = re.search("([^A-Za-z0-9\._:-])", bucket, re.UNICODE)
+        if invalid:
+            raise S3.Exceptions.ParameterError("Bucket name '%s' contains disallowed character '%s'. The only supported ones are: us-ascii letters (a-z, A-Z), digits (0-9), dot (.), hyphen (-), colon (:), and underscore (_)." % (bucket, invalid.groups()[0]))
+    elif dns_strict:
+        # This is the default
+        max_length = 63
         invalid = re.search("([^a-z0-9\.-])", bucket, re.UNICODE)
         if invalid:
             raise S3.Exceptions.ParameterError("Bucket name '%s' contains disallowed character '%s'. The only supported ones are: lowercase us-ascii letters (a-z), digits (0-9), dot (.) and hyphen (-)." % (bucket, invalid.groups()[0]))
```
fviard, line 263: Maybe you can profit of your PR to change the following in this file:
`import S3.Exceptions` into `from S3.Exceptions import ParameterError`

And so simplify all of its usage here.

S3/Utils.py
```diff
@@ -234,8 +234,30 @@ def time_to_epoch(t):
     raise S3.Exceptions.ParameterError('Unable to convert %r to an epoch time. Pass an epoch time. Try `date -d \'now + 1 year\' +%%s` (shell) or time.mktime (Python).' % t)
 
 
-def check_bucket_name(bucket, dns_strict=True):
-    if dns_strict:
+def check_bucket_name(bucket, dns_strict=True, name_quirks=False):
```
fviard, line 237: Now that us-east-1 also doesn't support the previous set of extended chars, I think that it is ok to not add your "name_quirks" and just add the ":" inside the char list checked when not in "dns_strict" mode.

btw, can you escape the "-" inside the regexp, to be able to have it before the additional characters, and also avoid errors in the future?

S3/Config.py
```diff
@@ -133,6 +133,8 @@ class Config(object):
     force = False
     server_side_encryption = False
     enable = None
+    # Used to allow colons in bucket names for Ceph compatibility
+    bucket_name_quirks = False
```
fviard, line 137: I don't like so much "quirks" and to restrict that to colons and Ceph.

Maybe you can call that something like: "bucket_name_extended",  "extended_bucket_name", "bucket_name_relaxed", "bucket_name_lenient".

or invert the condition like:
"bucket_name_strict = True", 

And in the comment say that it relax rules for bucket name for S3 compatible services that supports path style requests.

S3/S3.py
```diff
@@ -407,17 +407,21 @@ def bucket_create(self, bucket, bucket_location = None, extra_headers = None):
...
-            body += bucket_location
-            body += "</LocationConstraint></CreateBucketConfiguration>"
-            debug("bucket_location: " + body)
-            check_bucket_name(bucket, dns_strict = True)
+        if self.config.bucket_name_quirks:
+            # We are explicitly not AWS
+            check_bucket_name(bucket, dns_strict = False, name_quirks = True)
         else:
-            check_bucket_name(bucket, dns_strict = False)
+            if bucket_location:
```
fviard, line 414: The CreateBucketConfiguration should be set in all cases, that we use "name quirks" or not.

For the moment, keep the existing code but remove the "check_bucket_name" of this block.
(In the future, I will change the EU/US thing to avoid some specific issues, but it can be done after).

After the previous block, add a new block of code:
```
if self.config.bucket_name_quirks:
    checkbucket(dns_strict = False)
else:
   check_bucket(dns_strict= True)
```

snosratiershad:
@blkmajik, Are you still interested in this PR?

PR #1282, opened 2022-10-11
Enable pip cache in actions/setup-python

drakulavich:
It will create cache for pip packages based on:
* Runner OS
* Python version
* requirements.txt hash

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,6 +16,7 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+          cache: 'pip'
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,6 +21,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -r requirements.txt
+          pip install .
       - name: Cache minio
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -16,10 +16,11 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+          cache: 'pip'
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install .
+          pip install -r requirements.txt
```
fviard, line 23: This is good to cleanup s3cmd, but maybe you could still keep an additional line with:
`pip install .`
Because it will call the setup.py of s3cmd and ensure that it is working well also.

drakulavich: Sure, I agree. 
Updated

.github/workflows/test.yml
```diff
@@ -16,10 +16,12 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+          cache: 'pip'
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
           pip install .
+          pip install -r requirements.txt
```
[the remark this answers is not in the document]
drakulavich: Thank you! Totally agree.
Fixed

fviard:
Merged, thank you!

PR #1283, opened 2022-10-21
Support bucket/object tagging command

### LINKED ISSUE #911: Add object tagging feature
Would be very useful during `put`, `sync`, etc..

http://docs.aws.amazon.com/AmazonS3/latest/dev/object-tagging.html

jeffwang0516:
Support Put/Get/DeleteBucketTagging & Put/Get/DeleteObjectTagging ( resolve #911 )

AWS Spec Ref
- https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketTagging.html
- https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObjectTagging.html

### DIFF
--- a/S3/Crypto.py
+++ b/S3/Crypto.py
@@ -85,9 +85,9 @@ def sign_request_v2(method='GET', canonical_uri='/', params=None, cur_headers=No
     # valid sub-resources to be included in sign v2:
     SUBRESOURCES_TO_INCLUDE = ['acl', 'lifecycle', 'location', 'logging',
                                'notification', 'partNumber', 'policy',
-                               'requestPayment', 'torrent', 'uploadId',
-                               'uploads', 'versionId', 'versioning',
-                               'versions', 'website',
+                               'requestPayment', 'tagging', 'torrent',
+                               'uploadId', 'uploads', 'versionId',
+                               'versioning', 'versions', 'website',
                                # Missing of aws s3 doc but needed
                                'delete', 'cors', 'restore']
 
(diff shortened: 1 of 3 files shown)


fviard:
Looks nice, thank you.
I will try to review that fast!

fviard:
So, after looking at this, I think that it would be nice to be able to directly take the "tag" values for `settagging` and have a pretty print for `gettagging` instead of having xml input/outputs.

Here the xml is kind of just the request content, but there is no complex logic or customization that could request to expose that to the user.

What I would think would be something like that:
`s3cmd settagging s3://mybucket/myfile "MyFirstTag=value1&OtherKey=value2"

```
s3cmd gettagging s3://mybucket/myfile
s3://mybucket/myfile (object):
     MyfirstTag: value1
     OtherKey: value2
```

The & is mostly not used for tags, so could be an ok separator I think, but it requires to use quotes otherwise "the shell" might interpret it. And I'm thinking better to not separate by space as different argument each tag, because it might be good, in the future, to allow to have multiple files set in one time.

What do you think about that?

(btw the way, not for now, but for the future, I'm thinking about adding --iformat  and --oformat flags to be able to specify different formats for input or outputs.
for example, for input: "inline, xml", for output: "pretty,xml,csv,json".

jeffwang0516:
@fviard It looks better! I've updated the code.

fviard:
Finally merged. Sorry for the delay and thank you for the feature :-)

PR #1284, opened 2022-10-26
Support versioning configuration

### LINKED ISSUE #937: Add bucket operations to enable and disable versioning
It would be useful and consistent if s3cmd supported enabling of "versioning" on a bucket (http://docs.aws.amazon.com/AmazonS3/latest/dev/Versioning.html), like it does other configuration options (e.g. `setpolicy` and `accesslog`).
This is related to https://github.com/s3tools/s3cmd/issues/368.

jeffwang0516:
- Command added
  - ~Enable versioning: `s3cmd setversioning s3://mybucket`~
  - ~Suspend versioning: `s3cmd setversioning s3://mybucket --versioning-suspended`~
  - Enable versioning: `s3cmd setversioning s3://mybucket enable`
  - Suspend versioning: `s3cmd setversioning s3://mybucket disable`
- Add versioning status to command `info`
  ```
  $ s3cmd info s3://mybucket
  ...
  Location:  us-east-1
  Payer:     BucketOwner
  Versioning:Suspended
  Expiration Rule: none
  Policy:    none
  ...
  ```

Resolve #937

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -481,7 +481,10 @@ class S3(object):
             response['requester-pays'] = self.get_bucket_requester_pays(uri)
         except S3Error as e:
             response['requester-pays'] = None
-        response['versioning'] = self.get_versioning(uri)
+        try:
+            response['versioning'] = self.get_versioning(uri)
+        except S3Error as e:
+            response['versioning'] = None
         return response
 
     def website_info(self, uri, bucket_location = None):
@@ -1065,12 +1068,12 @@ class S3(object):
         response = self.send_request(request)
         return response
 
-    def set_versioning(self, uri, status):
+    def set_versioning(self, uri, enabled):
         headers = SortedDict(ignore_case = True)
-        headers['content-type'] = 'application/xml'
-        body = '<VersioningConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">\n'
-        body += '<Status>%s</Status>\n' % status
-        body += '</VersioningConfiguration>\n'
+        status = "Enabled" if enabled is True else "Suspended"
+        body = '<VersioningConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">'
+        body += '<Status>%s</Status>' % status
+        body += '</VersioningConfiguration>'
         debug(u"set_versioning(%s)" % body)
         headers['content-md5'] = compute_content_md5(body)
         request = self.create_request("BUCKET_CREATE", uri = uri,
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -481,6 +481,7 @@ class S3(object):
             response['requester-pays'] = self.get_bucket_requester_pays(uri)
         except S3Error as e:
             response['requester-pays'] = None
+        response['versioning'] = self.get_versioning(uri)
         return response
 
     def website_info(self, uri, bucket_location = None):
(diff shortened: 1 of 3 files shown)


S3/S3.py
```diff
@@ -481,6 +481,7 @@ def bucket_info(self, uri):
             response['requester-pays'] = self.get_bucket_requester_pays(uri)
         except S3Error as e:
             response['requester-pays'] = None
+        response['versioning'] = self.get_versioning(uri)
```
fviard, line 484: I think that you can put a try/except around that also, as I easily guess that not all servers support versioning

S3/S3.py
```diff
@@ -1064,6 +1065,27 @@ def set_acl(self, uri, acl):
         response = self.send_request(request)
         return response
 
+    def set_versioning(self, uri, status):
+        headers = SortedDict(ignore_case = True)
+        headers['content-type'] = 'application/xml'
```
fviard, line 1070: I don't think that you should set that.
We don't set it usually and it is not an issue. better to stay consistent.

S3/S3.py
```diff
@@ -1064,6 +1065,27 @@ def set_acl(self, uri, acl):
...
+        body = '<VersioningConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">\n'
+        body += '<Status>%s</Status>\n' % status
```
fviard, line 1072: No need to add the line breaks (\n) in the xml. This is a message for the server, not for a human.

S3/S3.py
```diff
@@ -1064,6 +1065,27 @@ def set_acl(self, uri, acl):
         response = self.send_request(request)
         return response
 
+    def set_versioning(self, uri, status):
```
fviard, line 1068: Instead of taking the status as string directly here, it is probably better to only take an "on/off" value.

fviard:
Thank you for your contribution, this is nice to have but I did a few reviews.

fviard:
Also, I don't like so much the following syntax to disable:
`Suspend versioning: s3cmd setversioning s3://mybucket --versioning-suspended`

I think that it is a little bit complicated for someone not knowing it already.

I would have prefer something like one of the following:
A) 
Enable: `s3cmd setversioning s3://mybucket`
Disable: `s3cmd delversioning s3://mybucket`

B)
Enable: `s3cmd setversioning s3://mybucket` enable
Disable: `s3cmd setversioning s3://mybucket` disable

s3cmd
```diff
@@ -2047,6 +2049,20 @@ def cmd_setacl(args):
         update_acl(s3, uri, seq_label)
     return EX_OK
 
+def cmd_setversioning(args):
+    cfg = Config()
+    s3 = S3(cfg)
+    uri = S3Uri(args[0])
```
fviard, line 2055: This should work only on a bucket, so I think that it would be nice to add a check to avoid mistakes/confusions:
```
    if bucket_uri.object():
        raise ParameterError("Only bucket name is required for [setversioning] command")
```

jeffwang0516:
@fviard Thanks for the reviews! I've pushed my changes.

fviard:
It looks good now, thanks for the changes.
I have merged it :-)

fviard:
To be noted for a futur new PR, in such a case, you should "squash" your commits or have relevant commit messages.
I did not take care and so the merge added 2 commits: the original one, and the "refactoring" commit.
It is not too bad, so i don't fix that with a force push, but would be nice to take care of that in the future to have a nice git history :-)

jeffwang0516:
Got it, thanks!

PR #1290, opened 2022-11-10
Use generate_content_md5() in set_versioning command

jeffwang0516:
Fix to pr #1284, 
I didn't notice that there's a change from `compute_content_md5` to `generate_content_md5`.


fviard:
Thank you but I merged another PR that fixed the same issue:
https://github.com/s3tools/s3cmd/pull/1340

PR #1295, opened 2022-11-25
Avoid unrecognized escape sequences

kraai:
https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals
indicates that unrecognized escape sequences produce a
`DeprecationWarning` and will eventually be a `SyntaxError`, so avoid
them.

### DIFF
--- a/S3/BaseUtils.py
+++ b/S3/BaseUtils.py
@@ -74,8 +74,8 @@ __all__.append("md5")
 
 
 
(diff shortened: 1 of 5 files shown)


fviard:
Thank you for this PR.
Sadly I missed it and I did again the change while fixing something else, so I will close it. But it would have been good otherwise.
I will try to review a lot faster any further PR that would like to do in the future as they were still worthwhile.

PR #1296, opened 2022-11-25
Pass --include to curl

kraai:
`run-tests.py` passes `-include` to curl, which curl interprets as
`--include --netrc --cookie-jar lude`.  Pass `--include` instead.


fviard:
That is a good catch, thank you!

PR #1297, opened 2022-12-09
Add setobjectretention and setobjectlegalhold commands

withings-sas:
This time the XML is generated by s3cmd


fviard:
Very good like that, merged, thank you very much and sorry for the long delay.

Issue #1313, opened 2023-03-14
"TypeError: string indices must be integers" in dry-run


xcir:
An error occurred during the dry run execution.

# My Environment

OS: Ubuntu 20.04.4(focal)
Python: 3.8.10
s3cmd rev: [6f3e1ba](https://github.com/s3tools/s3cmd/commit/6f3e1baa667da53422f60bd941112e1f3a07662c)

# Traceback

```
Invoked as: ./s3cmd sync --follow-symlinks --files-from=- -c /**** --list-allow-unordered --dry-run /**** s3://****
Problem: <class 'TypeError: string indices must be integers
S3cmd:   2.3.0-dev
python:   3.8.10 (default, Jun 22 2022, 20:18:18)
[GCC 9.4.0]
environment LANG=C

Traceback (most recent call last):
  File "./s3cmd", line 3461, in <module>
    rc = main()
  File "./s3cmd", line 3358, in main
    rc = cmd_func(args)
  File "./s3cmd", line 2149, in cmd_sync
    return cmd_sync_local2remote(args)
  File "./s3cmd", line 2116, in cmd_sync_local2remote
    ret = _single_process(args[:-1])
  File "./s3cmd", line 1863, in _single_process
    return _child(destination_base, source_args)
  File "./s3cmd", line 2012, in _child
    output(u"remote copy: '%s' -> '%s'" % (relative_file['copy_src'], relative_file))
TypeError: string indices must be integers
```

# Possible Solution

https://github.com/s3tools/s3cmd/blob/f8eafbb5a8ed95268dbdadbf6df49f2a4c93393e/s3cmd#L2011-L2012

Maybe "item['copy_src']" is correct.

fviard:
Thanks for the detailed report in addition of the fix :-)

PR #1314, opened 2023-03-14
Fix type error(#1313)

xcir:
Fix to #1313


fviard:
Good catch, merged, thank you!

PR #1350, opened 2023-11-09
Update S3.py

visbits:


fviard:
I'm not sure that the fix like that would be a good thing.
If you do that, it means that whatever will be your command, you will have all the subkeys when you use allow-unordered, like if you enabled recursive.
That could be unexpected and have side effect. I think, no?

Maybe a solution for this case, that would be a little bit more painful but safer, is to refuse commands that when you have the incompatible set "not recursive" and "allow-unordered" at the same time?

PR #1353, opened 2023-11-13
Provide sha256 checksum with file updates

### LINKED ISSUE #1177: Missing Content-MD5 http header for put request on object with bucket defined Lock
When an aws s3 bucket is configured with Object Lock in order to benefit from WORM feature the upload
cf: https://aws.amazon.com/fr/blogs/storage/protecting-data-with-amazon-s3-object-lock/

the upload fails with the following error:

`Content-MD5 HTTP header is required for Put Object requests with Object Lock parameters`

Is there a way add Content-MD5 header to the request ?

I could have used `--add-header=NAME:VALUE` but since I'm using built in gpg encryption, I can't compute the md5 before running the command.

Any idea ?

eullerborges:
Fixes #1177. I tried `Content-MD5` as well, but that was more involved.

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -11,6 +11,7 @@
 ## --------------------------------------------------------------------
 
 from __future__ import absolute_import, division
+import base64
 
 import sys
 import os
@@ -1724,6 +1725,10 @@ class S3(object):
         else:
             sha256_hash = checksum_sha256_file(stream, offset, size_total)
         request.body = sha256_hash
+
+        # Provide the checksum with teh request. This is important for buckets that have 
+        # Object Lock enabled.
+        headers['x-amz-checksum-sha256'] = base64.b64encode(sha256_hash.digest()).decode()
 
         if use_expect_continue:
             if not size_total:


S3/S3.py
```diff
@@ -1725,6 +1726,10 @@ def send_file(self, request, stream, labels, buffer = '', throttle = 0,
             sha256_hash = checksum_sha256_file(stream, offset, size_total)
         request.body = sha256_hash
 
+        # Provide the checksum with teh request. This is important for buckets that have
```
rampageservices, line 1729: "the" is spelled incorrectly here, shown as "teh". Minor typo.

S3/S3.py
```diff
@@ -1725,6 +1726,10 @@ def send_file(self, request, stream, labels, buffer = '', throttle = 0,
             sha256_hash = checksum_sha256_file(stream, offset, size_total)
         request.body = sha256_hash
 
+        # Provide the checksum with teh request. This is important for buckets that have 
+        # Object Lock enabled.
+        headers['x-amz-checksum-sha256'] = base64.b64encode(sha256_hash.digest()).decode()
```
rampageservices, line 1731: `base64.b64encode(sha256_hash.digest()).decode()`
Can this line be broken down? It is doing a lot all at once. It is hard for readability.
Perhaps we  canpull out the digest operation into a separate sha256_hash_digest variable?
Or maybe this functionality can be separated into a new function in S3/Crypto.py

I know this exact line is recommended online in tutorials but is in my opinion hard to read.
See: https://debugpointer.com/python/sha256-to-base64-python

rampageservices (changes requested):
Typo needs fixing.
Suggest improvement for readability.

S3/S3.py
snosratiershad, line 1731: As someone could use s3cmd as library this line can override the header potentially comes with request argument's headers attribute. I think it's better to check if it currently has values:
```python
if not headers.get("x-amz-checksum-sha256"):
    headers['x-amz-checksum-sha256'] = sha256_hash_to_base64(sha256_hash)
```

Also in this module encodestring (which imported above as an alias of encodebytes when available) is the function used for base64 encoding.
I think you can use:
```python
sha256_hash_digest_b64 = encodestring(sha256_hash_digest).strip().decode()
```
to align with other usecases.

snosratiershad:
I think we can close this PR as #1393 has completed this reviews.

fviard:
Closing as suggested, to be replaced by #1393

eullerborges:
Sorry for not coming back to this, I've been away from GitHub for a while. I'm glad this was picked up and merged in. Thank you!

Issue #1363, opened 2024-01-16
Exception when running s3cmd sync


banjoh:
### Description

When copying files from one bucket to a local drive, I saw this exception

```
….
download: ‘s3://....' -> '/tmp/archives/....' (9422831 bytes in 0.0 seconds, 307.48 MB/s) [162 of 163]
download: 's3://....' -> '/tmp/archives/....' (9309 bytes in 0.0 seconds, 7.06 MB/s) [163 of 163]

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
    An unexpected error has occurred.
  Please try reproducing the error using
  the latest s3cmd code from the git master
  branch found at:
    https://github.com/s3tools/s3cmd
  and have a look at the known issues list:
    https://github.com/s3tools/s3cmd/wiki/Common-known-issues-and-their-solutions-(FAQ)
  If the error persists, please report the
  following lines (removing any private
  info as necessary) to:
   s3tools-bugs@lists.sourceforge.net


!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

Invoked as: /usr/bin/s3cmd --access_key=kurl --secret_key=<redacted> --host=minio.minio --no-ssl --host-bucket=<redacted> sync s3://<redacted_bucket> /tmp/archives
Problem: <class 'TypeError: 'dict_keys' object is not subscriptable
S3cmd:   2.4.0
python:   3.12.1 (main, Dec  8 2023, 02:59:13) [GCC 13.2.0]
environment LANG=NOTSET

Traceback (most recent call last):
  File "/usr/bin/s3cmd", line 3627, in <module>
    rc = main()
         ^^^^^^
  File "/usr/bin/s3cmd", line 3524, in main
    rc = cmd_func(args)
         ^^^^^^^^^^^^^^
  File "/usr/bin/s3cmd", line 2163, in cmd_sync
    return cmd_sync_remote2local(args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/bin/s3cmd", line 1719, in cmd_sync_remote2local
    _set_local_filename(copy_pairs, destination_base, source_args, dir_cache)
  File "/usr/bin/s3cmd", line 1408, in _set_local_filename
    source_args[0] == remote_list[remote_list.keys()[0]].get(u'object_uri_str', ''):
                                  ~~~~~~~~~~~~~~~~~~^^^
[...]

PR #1364, opened 2024-01-23
Fix `dict_keys` indexing

tdeni:
s3cmd fails on python3.6 as follows
```sh
  File "/usr/bin/s3cmd", line 3627, in <module>
    rc = main()
  File "/usr/bin/s3cmd", line 3524, in main
    rc = cmd_func(args)
  File "/usr/bin/s3cmd", line 2163, in cmd_sync
    return cmd_sync_remote2local(args)
  File "/usr/bin/s3cmd", line 1719, in cmd_sync_remote2local
    _set_local_filename(copy_pairs, destination_base, source_args, dir_cache)
  File "/usr/bin/s3cmd", line 1408, in _set_local_filename
    source_args[0] == remote_list[remote_list.keys()[0]].get(u'object_uri_str', ''):
TypeError: 'dict_keys' object does not support indexing
```

Fixes:
- #1363

### DIFF
--- a/s3cmd
+++ b/s3cmd
@@ -1405,7 +1405,7 @@ def cmd_sync_remote2local(args):
             if not os.path.isdir(deunicodise(destination_base)):
                 raise ParameterError("Destination is not an existing directory")
         elif len(remote_list) == 1 and \
-             source_args[0] == remote_list[remote_list.keys()[0]].get(u'object_uri_str', ''):
+            source_args[0] == remote_list[list(remote_list.keys())[0]].get(u'object_uri_str', ''):
             if os.path.isdir(deunicodise(destination_base)):
                 raise ParameterError("Destination already exists and is a directory")
             remote_list[remote_list.keys()[0]]['local_filename'] = destination_base


snosratiershad:
I think we don't currently support 3.6, I'll research for the version that indexing support added.

snosratiershad:
It also fails in recent versions of python with:
```
TypeError: 'dict_keys' object is not subscriptable
```

The change is required. Thanks @tdeni LGTM.

snosratiershad:
@fviard, at first I was wrongly assumed that remote_lists is always a dict, which directed my to test the dict_values instead of the code.
After your message I've made a deeper investigation:
the code mentioned:
```python
source_args[0] == remote_list[remote_list.keys()[0]].get(u'object_uri_str', ''):
```
is inside of this function:
```python
def _set_local_filename(remote_list, destination_base, source_args, dir_cache):
```
the function called 3 times in the code:
```python
_set_local_filename(remote_list, destination_base, source_args, dir_cache)
_set_local_filename(update_list, destination_base, source_args, dir_cache)
_set_local_filename(copy_pairs, destination_base, source_args, dir_cache)
```
I've checked type of three of `remote_list`, `update_list` and `copy_pairs`.`remote_list` and `update_list` are `FileDict`s, which inherited from `SortedDict` which has custom `keys` function returning `list` not `dict_keys`, but the last one `copy_pairs` has `dict` type and could cause the bug.
copy_pairs comes from return of `compare_filelists` which I think should return instance of SortedDict instead of dict to solve the issue instead of modifying the code that statement getting `object_uri_str` itself.
To have values in `copy_pairs`, their should be a file exists in the dest with same md5 of a file need to be synced and not same with it's relevant file in the dest. This could be reproduced by creating one file with exact content of file to be synced in dest directory and it gives `copy_pairs` exactly length 1 value. but there is still another condition, this statement should not be satisfied:
```python
if destination_base.endswith(os.path.sep):
```
But the destination is checked to end with sep, so I couldn't reproduce it:
```
ERROR: Parameter problem: Destination must be a directory and end with '/' when downloading multiple sources.
```

snosratiershad:
I've commented the check:
```python
    if not destination_base.endswith(os.path.sep):
        if fetch_source_args[0].endswith(u'/') or len(fetch_source_args) > 1:
            raise ParameterError("Destination must be a directory and end with '/' when downloading multiple sources.")
```
and reproduced the bug:

image

fviard:
@snosratiershad thank you for your investigation, but on my side, even with your "steps to reproduce" i'm not able to reproduce the issue :'(

You can confirm me that you are on Linux and your configuration is the default one except for credentials?

snosratiershad:
@fviard Yes I'm on linux and with no specific configuration.
Just to make sure:
```
$ s3cmd sync s3://BUCKET/newdir /tmp/dest/
```
Did you noticed that there is no slash or separator after newdir?

snosratiershad:
@fviard I've removed my configuration to test again and noticed I've missed one step to reproduce. updated.

fviard:
I have just pushed a fix for this issue, similar but a little bit more efficient:
https://github.com/s3tools/s3cmd/commit/dae680c53ea6973ce88e531ae5a4a9dedc97123a

Thank you very much for your feedbacks that were invaluable to me for resolving this unexplained issue that haunted me for a long time.

Good job also @snosratiershad to have managed to notice that the copy_pairs was still a dict, and to have managed to give me a test case to reproduce.

Just for your info, there was still a small error in your test case for me being able to reproduce:
"s3cmd sync s3://BUCKET/newdir /tmp/dest"
I needed to not have a final "slash" for the destination.

So in the end, the key point was to have a copy listed in the copy_pair, but just a "single" one and also not have the destination with a trailing slash!

fviard:
Fixed with:
https://github.com/s3tools/s3cmd/commit/dae680c53ea6973ce88e531ae5a4a9dedc97123a

Issue #1027, opened 2019-01-03
AWS credentials environment variable has changed


edgewood:
s3cmd looks for the AWS_CREDENTIAL_FILE environment variable to find the AWS credentials in the absence of a s3cmd config file.  However, newer versions of awscli (1.7.45 and later) use a different environment variable, AWS_SHARED_CREDENTIALS_FILE  See 
https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html

As a workaround, export AWS_CREDENTIAL_FILE=$AWS_SHARED_CREDENTIALS_FILE works for me with a very simple AWS setup, but there could be other changes to the file structure in other setups.

As an aside, I discovered this issue trying to use AWS credentials as a workaround for s3cmd's lack of support for the XDG basedir spec (#658).

PR #1365, opened 2024-02-16
Fixes #1361 and #1027 - update to python 3.2+ API (existing code calls removed API)

### LINKED ISSUE #1361: obscure error message for .s3cfg not found
%  s3cmd --configure

  Please try reproducing the error using
  the latest s3cmd code from the git master
  branch found at:
    https://github.com/s3tools/s3cmd
  and have a look at the known issues list:
    https://github.com/s3tools/s3cmd/wiki/Common-known-issues-and-their-solutions-(FAQ)
  If the error persists, please report the
  following lines (removing any private
  info as necessary) to:
   s3tools-bugs@lists.sourceforge.net

Invoked as: /usr/local/bin/s3cmd --configure
Problem: <class 'AttributeError: 'ConfigParser' object has no attribute 'readfp'
S3cmd:   2.4.0
python:   3.12.1 (main, Dec  8 2023, 19:21:52) [Clang 14.0.3 (clang-1403.0.22.14.1)]
environment LANG=en_US.UTF-8

  File "/usr/local/lib/python3.12/site-packages/S3/Config.py", line 269, in __init__
    self.read_config_file(configfile)
  File "/usr/local/lib/python3.12/site-packages/S3/Config.py", line 551, in read_config_file
    cp = ConfigParser(configfile)
         ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/S3/Config.py", line 640, in __init__
    self.parse_file(file, sections)
  File "/usr/local/lib/python3.12/site-packages/S3/Config.py", line 652, in parse_file
    with io.open(file, "r", encoding=self.get('encoding', 'UTF-8')) as fp:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/Users/michaelreiche/.s3cfg'

During handling of the above exception, another exception occurred:

  File "/usr/local/bin/s3cmd", line 3627, in <module>
    rc = main()
         ^^^^^^
  File "/usr/local/bin/s3cmd", line 3313, in main
    cfg = Config(options.config, options.access_key, options.secret_key, options.access_token)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/S3/Config.py", line 272, in __init__
    self.aws_credential_file()
  File "/usr/local/lib/python3.12/site-packages/S3/Config.py", line 459, in aws_credential_file
    config.readfp(io.StringIO(config_string))
    ^^^^^^^^^^^^^
AttributeError: 'ConfigParser' object has no attribute 'readfp'. Did you mean: 'read'?

  Please try reproducing the error using
  the latest s3cmd code from the git master
  branch found at:
    https://github.com/s3tools/s3cmd
  and have a look at the known issues list:
    https://github.com/s3tools/s3cmd/wiki/Common-known-issues-and-their-solutions-(FAQ)
  If the error persists, please report the
  above lines (removing any private
  info as necessary) to:
   s3tools-bugs@lists.sourceforge.net

ms-jpq:
Fixes:

- https://github.com/s3tools/s3cmd/issues/1361
- https://github.com/s3tools/s3cmd/issues/1027

Should be OK to replace the removed API call, since the replacement was added in python 3.2.

https://docs.python.org/3/library/configparser.html

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -73,7 +73,7 @@ except NameError:
     # In python 3, unicode -> str, and str -> bytes
     unicode = str
 
-PY32 = (sys.version_info >= (3, 2))
+PY3 = (sys.version_info >= (3, 0))
 
 
 def is_bool_true(value):
@@ -459,7 +459,7 @@ class Config(object):
             try:
                 try:
                     buf = io.StringIO(config_string)
-                    if PY32:
+                    if PY3:
                       config.read_file(buf)
                     else:
                       config.readfp(buf)
@@ -471,7 +471,7 @@ class Config(object):
                     # to be able to read the file with PyConfigParser()
                     config_string = u'[default]\n' + config_string
                     buf = io.StringIO(config_string)
-                    if PY32:
+                    if PY3:
                       config.read_file(buf)
                     else:
                       config.readfp(buf)
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -268,7 +268,7 @@ class Config(object):
             try:
                 self.read_config_file(configfile)
             except IOError:
-                if 'AWS_CREDENTIAL_FILE' in os.environ or 'AWS_PROFILE' in os.environ:
+                if 'AWS_SHARED_CREDENTIALS_FILE' in os.environ or 'AWS_CREDENTIAL_FILE' in os.environ or 'AWS_PROFILE' in os.environ:
                     self.aws_credential_file()
 
             # override these if passed on the command-line
@@ -439,7 +439,8 @@ class Config(object):
     def aws_credential_file(self):
         try:
             aws_credential_file = os.path.expanduser('~/.aws/credentials')
-            credential_file_from_env = os.environ.get('AWS_CREDENTIAL_FILE')
+            credential_file_from_env = os.environ.get('AWS_SHARED_CREDENTIALS_FILE') \
+                or os.environ.get('AWS_CREDENTIAL_FILE')
             if credential_file_from_env and \
                os.path.isfile(credential_file_from_env):
                 aws_credential_file = base_unicodise(credential_file_from_env)
@@ -454,9 +455,7 @@ class Config(object):
                 config_string = fp.read()
             try:
                 try:
-                    # readfp is replaced by read_file in python3,
-                    # but so far readfp it is still available.
-                    config.readfp(io.StringIO(config_string))
+                    config.read_file(io.StringIO(config_string))
                 except MissingSectionHeaderError:
                     # if header is missing, this could be deprecated
                     # credentials file format as described here:
@@ -464,7 +463,7 @@ class Config(object):
                     # then do the hacky-hack and add default header
                     # to be able to read the file with PyConfigParser()
                     config_string = u'[default]\n' + config_string
-                    config.readfp(io.StringIO(config_string))
+                    config.read_file(io.StringIO(config_string))
             except ParsingError as exc:
                 raise ValueError(
                     "Error reading aws_credential_file "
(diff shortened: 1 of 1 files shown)


TurtleWilly:
> Should be OK to replace the removed API call, since the replacement was added in python 3.2.

Wouldn't that break Python 2.x support?

I still see tons of legacy Python 2 fallback code all over the place in s3cmd's code. IMHO removing Python 2 support should be done in a more organized fashion (aka all at once, then bump the major version) rather than a tidbit here or there and breaking things for people that still expect Python 2 support to work as the documentation suggests: "S3cmd requires Python 2.6 or newer." (quote from master readme.md). _[Note: I use 3.12, but I know how it is when projects/ packages break backwards compatibility for no particular reasons. ]_

lavigne958:
> > Should be OK to replace the removed API call, since the replacement was added in python 3.2.
> 
> Wouldn't that break Python 2.x support?
> 
> I still see tons of legacy Python 2 fallback code all over the place in s3cmd's code. IMHO removing Python 2 support should be done in a more organized fashion (aka all at once, then bump the major version) rather than a tidbit here or there and breaking things for people that still expect Python 2 support to work as the documentation suggests: "S3cmd requires Python 2.6 or newer." (quote from master readme.md). _[Note: I use 3.12, but I know how it is when projects/ packages break backwards compatibility for no particular reasons. ]_

I agree with you this will break code for people using python2.X

though if nothing is done, it will break code for people using python3.12+ so either end it will break someone's code.

So far python2.7 has been deprecated in 2008, and reached its end-of-life in 2020. I think it's time to move forward.
Users with the strong need to support python2 can stay on current release of s3cmd. 

I believe things should be done in this order:

1. remove the mention "requires python2.6" from the README 
2. update the code to use `config.read_file()`

Later on more changes could be added to remove python2 support.

fviard:
Thank you for your contributions.
Sorry for the very long feedback, I would prefer to still not break python 2.x support when backward compatibility is still not difficult to preserve.

In the current case, it would be easy to fix like it is done in other files:
- Adding :
` PY3 = (sys.version_info >= (3, 0))` at the top of this source file.
- Using a : 
  ```
  if PY3:
        ...read_file...
  else:
         ...readfp...
  ```

Would it be possible for you to fix your PR with that? Otherwise, as I know that it has been a long time, I will do it :-)

ms-jpq:
> Thank you for your contributions. Sorry for the very long feedback, I would prefer to still not break python 2.x support when backward compatibility is still not difficult to preserve.
> 
> In the current case, it would be easy to fix like it is done in other files:
> 
>     * Adding :
>       ` PY3 = (sys.version_info >= (3, 0))` at the top of this source file.
> 
>     * Using a :
>       ```
>       if PY3:
>             ...read_file...
>       else:
>              ...readfp...
>       ```
> 
> 
> Would it be possible for you to fix your PR with that? Otherwise, as I know that it has been a long time, I will do it :-)

thanks 👍

S3/Config.py
```diff
@@ -73,6 +73,8 @@
     # In python 3, unicode -> str, and str -> bytes
     unicode = str
 
+PY32 = (sys.version_info >= (3, 2))
```
fviard, line 76: PY3 will be enough (and consistent with what we use everywhere else) :-) 
(We are probably not compatible with python 3.0 to 3.4 that were not really a serious thing anyway)

ms-jpq: ty, pls squash since i made a silly commit message 👍

fviard:
Merged, thank you very much!

PR #1393, opened 2024-08-22
Added x-amz-checksum-sha256 header for Object Lock support

### LINKED ISSUE #1177: Missing Content-MD5 http header for put request on object with bucket defined Lock

AlexisPPLIN:
This is a follow-up of PR #1353 which is inactive since Dec 5, 2023

As asked I moved the sha256 digest to base64 string into a separate function into `S3/Crypto.py`

Fixes #1177


fviard:
Sorry for the long delay of my reply.
As you can see with the failed tests, you forgot the import for base64 in crypto.
In fact, as it is done for encodestring, you can directly import the function that you want to use.

AlexisPPLIN:
Hi !
No worries for the delay

I fixed my import like you advised me to.
I also rebased my PR on master.

S3/Crypto.py
```diff
@@ -346,3 +348,11 @@ def calculateChecksum(buffer, mfile, offset, chunk_size, send_chunk):
 
     return md5_hash.hexdigest()
 __all__.append("calculateChecksum")
+
+def sha256_hash_to_base64(sha256_hash):
+    # Extract digest from sha256 hash
+    sha256_hash_digest = sha256_hash.digest()
+    # Then convert it to base64
+    sha256_hash_digest_b64 = b64encode(sha256_hash_digest).decode()
```
snosratiershad, line 356: In this module encodestring (which imported above as an alias of encodebytes when available) is the function used for base64 encoding.
I think you can use:
```python
sha256_hash_digest_b64 = encodestring(sha256_hash_digest).strip().decode()
```
to align with other usecases.

S3/S3.py
```diff
@@ -1848,6 +1848,11 @@ def send_file(self, request, stream, labels, buffer = '', throttle = 0,
             sha256_hash = checksum_sha256_file(stream, offset, size_total)
         request.body = sha256_hash
 
+        # Provide the checksum with the request. This is important for buckets that have 
+        # Object Lock enabled.
+
+        headers['x-amz-checksum-sha256'] = sha256_hash_to_base64(sha256_hash)
```
snosratiershad, line 1854: As someone could use s3cmd as library this line can override the header potentially comes with request argument's headers attribute. I think it's better to check if it's currently has values:
```python
if not headers.get("x-amz-checksum-sha256"):
    headers['x-amz-checksum-sha256'] = sha256_hash_to_base64(sha256_hash)
```

PR #1402, opened 2024-10-30
Environment Variable For STS Endpoint 

ahmadalkhansa:
By having modifiable environment variable, it is possible to perform authentication with storage solutions that implement Security Token Service (STS) and S3 API.

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -323,7 +323,7 @@ class Config(object):
                 web_identity_token_file = os.environ.get('AWS_WEB_IDENTITY_TOKEN_FILE')
                 if web_identity_token_file:
                     with open(web_identity_token_file) as f:
-                        web_identity_token = f.read()
+                        web_identity_token = f.read().rstrip()
                     params['Action'] = 'AssumeRoleWithWebIdentity'
                     params['WebIdentityToken'] = web_identity_token
                 encoded_params = '&'.join([
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -330,7 +330,9 @@ class Config(object):
                     '%s=%s' % (k, s3_quote(v, unicode_output=True))
                     for k, v in params.items()
                 ])
-                sts_endpoint = "sts.amazonaws.com"
+                sts_endpoint = os.environ.get("AWS_STS_ENDPOINT") 
+                if sts_endpoint == None:
+                    sts_endpoint = "sts.amazonaws.com"
                 if os.environ.get("AWS_STS_REGIONAL_ENDPOINTS") == "regional":
                     # Check if the AWS_REGION variable is available to use as a region.
                     region = os.environ.get("AWS_REGION")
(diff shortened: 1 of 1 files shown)


S3/Config.py
```diff
@@ -323,14 +323,16 @@ def role_config(self):
...
-                        web_identity_token = f.read()
+                        web_identity_token = f.read().rstrip()
                     params['Action'] = 'AssumeRoleWithWebIdentity'
                     params['WebIdentityToken'] = web_identity_token
                 encoded_params = '&'.join([
                     '%s=%s' % (k, s3_quote(v, unicode_output=True))
                     for k, v in params.items()
                 ])
-                sts_endpoint = "sts.amazonaws.com"
+                sts_endpoint = os.environ.get("AWS_STS_ENDPOINT")
```
snosratiershad, line 333: You can provide default value directly to `environ.get`:
```python
sts_endpoint = os.environ.get("AWS_STS_ENDPOINT", default="sts.amazonaws.com")
```

snosratiershad:
@ahmadalkhansa, How do you think about adding `sts_endpoint` also to config file variables?

ahmadalkhansa:
Hi, thank you for your feedback! I've modified `environ.get` and added `sts_endpoint` to config file variables as requested.

S3/Config.py
```diff
@@ -324,14 +326,15 @@ def role_config(self):
...
-                        web_identity_token = f.read()
+                        web_identity_token = f.read().rstrip()
                     params['Action'] = 'AssumeRoleWithWebIdentity'
                     params['WebIdentityToken'] = web_identity_token
                 encoded_params = '&'.join([
                     '%s=%s' % (k, s3_quote(v, unicode_output=True))
                     for k, v in params.items()
                 ])
-                sts_endpoint = "sts.amazonaws.com"
+                if len(sts_endpoint) == 0:
```
fviard, line 336: nitpick but you can simply do: `if not sts_endpoint:`

S3/Config.py
```diff
@@ -324,14 +326,15 @@ def role_config(self):
...
+                    sts_endpoint = os.environ.get("AWS_STS_ENDPOINT", default="sts.amazonaws.com")
```
fviard, line 337: In fact, usually "environment" vars have the higher priority to override the config, so it would be better that it comes first.
So, probably, if i'm not wrong, you can put "sts.amazonaws.com" directly as the default config in the config file.
So then you will not need your "if" anymore:
`sts_endpoint = os.environ.get("AWS_STS_ENDPOINT", sts_endpoint)`

(To be noted, you don't need to put the "default=" because it is the single and usual second argument for such getters)

ahmadalkhansa:
Thank you! I think the last commit should provide precendence to the environment variable without the need for a if statement.

fviard:
thank you for the change, merged!

PR #1408, opened 2024-12-18
Fix file extension detection for files with multiple dot separators

nerlihmax:
Hi!

I have issues with the flag `--add-encoding-exts=js,html,css` when uploads js static on production buckets. My files have names like `widget-no-csr.oWiXNN3Poy.js` or `widget-no-csr.oWiXNN3Poy.js.gz`, where is more than one dot symbol separator. Because https://github.com/s3tools/s3cmd/blob/10aea0918082ea87723a7e6754133283883bcfa7/S3/S3.py#L736 there is file extension `oWiXNN3Poy` for first file and `js` for second. That looks like a bug. I fixed that, but I think it could be kind of dangerous because it breaks backward compability.

### DIFF
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -738,7 +738,7 @@ class S3(object):
         parts = filename.rsplit('.', 1)
         if len(parts) < 2:
             return False
-        ext = parts[-1]
+        ext = parts[1]
         if ext in exts:
             return True
         else:
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -728,15 +728,14 @@ class S3(object):
         return content_type
 
     def add_encoding(self, filename, content_type):
+        if not filename:
+            return False
         if 'charset=' in content_type:
            return False
         exts = self.config.add_encoding_exts.split(',')
         if exts[0]=='':
             return False
-        parts = filename.rsplit('.',2)
-        if len(parts) < 2:
-            return False
-        ext = parts[1]
+        ext = filename.rsplit('.', 1)[-1]
         if ext in exts:
             return True
         else:
--- a/S3/S3.py
+++ b/S3/S3.py
@@ -735,7 +735,10 @@ class S3(object):
         exts = self.config.add_encoding_exts.split(',')
         if exts[0]=='':
             return False
-        ext = filename.rsplit('.', 1)[-1]
+        parts = filename.rsplit('.', 1)
+        if len(parts) < 2:
+            return False
+        ext = parts[-1]
         if ext in exts:
             return True
         else:


fviard:
Thank you for your PR, indeed the existing code looks very strange, and my best guess is that it was not done on purpose but a confusion with the count argument of the python split.

S3/S3.py
```diff
@@ -728,15 +728,14 @@ def content_type(self, filename=None, is_dir=False):
...
         if 'charset=' in content_type:
            return False
         exts = self.config.add_encoding_exts.split(',')
         if exts[0]=='':
             return False
-        parts = filename.rsplit('.',2)
-        if len(parts) < 2:
-            return False
-        ext = parts[1]
+        ext = filename.rsplit('.', 1)[-1]
```
fviard, line 738: If i'm not wrong, you should still add a check for the len of the split to ensure that it is at least of 2.
because otherwise a file juste named "js" (and not "something.js" or even ".js") would be considered as the extension, what I think would be wrong.

snosratiershad: @fviard is right, I've tested it.

snosratiershad: Actually the only thing wrong with current code is putting maxsplit 2 instead of 1. cause parts[1] is equal to parts[-1] if there is two index, and there can't be more then two index if maxsplit is 1.

nerlihmax: @fviard You are right, but it will be impossible to add encoding extensions for files with names without extensions, like `myfile`. Are you sure about it?

snosratiershad: @nerlihmax, nobody will put the name `myfile` in `add_encoding_exts` so it will be rejected below by default:
```python
        if ext in exts:
            return True
        else:
            return False
```

S3/S3.py
```diff
@@ -728,15 +728,17 @@ def content_type(self, filename=None, is_dir=False):
...
         if 'charset=' in content_type:
            return False
         exts = self.config.add_encoding_exts.split(',')
         if exts[0]=='':
             return False
-        parts = filename.rsplit('.',2)
+        parts = filename.rsplit('.', 1)
         if len(parts) < 2:
             return False
-        ext = parts[1]
```
snosratiershad, line 739: now we have exactly two parts here and parts[-1] equals to parts[1]. I think parts[1] is more simple.

snosratiershad:
@nerlihmax, @fviard LGTM.

fviard:
Merged, thank you.
@nerlihmax For info if you do other PR, it would be good to "clean" or "squash" the commits in your PR for them to be clean once merged.
Here I did with a "squash" through "github" but it is always better if the PR can be merged as-is :-)

PR #1410, opened 2025-01-01
fix: redirection based on location header for methods other than HEAD

snosratiershad:
The location header can be provided with methods other than HEAD.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ s3cmd.egg-info
 s3cmd.spec
 .idea
 .s3cfg
+.python-version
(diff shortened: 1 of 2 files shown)


fviard:
Could you give me a little bit more context of why it is needed or with which server you are encountering an issue that can motivate this change?

Because from my point of view, the kind of "redirection" that we are looking for are supposed to be "application data" that is in the body, and so, if we don't have a data and that it is not a HEAD request, it might probably mean that we are not targeting a wrong service or endpoint.

snosratiershad:
@fviard, redirection is not always initiated by a s3 system, it could be initiated by gateways, etc.
My case was trying to list my objects from ArvanCloud S3, and got redirection response with GET method. So I fixed it for myself using changes in this PR.

snosratiershad:
@fviard Hi, I found another use case in this time. A type of regular redirection in some of service providers is to for using subdomain based addressing, like redirecting from `s3serviceprovider.com/bucketname` to `bucketname.s3serviceprovider.com`

PR #1412, opened 2025-01-13
feat: warning when config file is world or group readable

### LINKED ISSUE #1405: Requiring ~/.s3cfg to be non-readable by non-owner
## Bug
s3cmd seems to permit the ~/.s3cfg file to have any permissions for reading by other users. Since the config file may contain the `secret_key`, it's important for this file to be readable only by the owner of the file to prevent unauthorized users from reading it.

## Suggested solution
`sshd` supports a `StrictModes` flag (default enabled) which refuses to work with a given ssh private key if the key file has wide open permissions.

Can s3cmd similarly default to refuse to load the ~/.s3cfg file if its permissions are too wide open?

snosratiershad:
Fixes #1405.


snosratiershad:
Closing as declined in #1405

PR #1420, opened 2025-04-03
Option to specify profile

railadeividas:
Allow having few profiles in default config .s3cfg

### DIFF
--- a/S3/Config.py
+++ b/S3/Config.py
@@ -261,18 +261,19 @@ class Config(object):
     max_retries = 5
 
     ## Creating a singleton
-    def __new__(self, configfile = None, access_key=None, secret_key=None, access_token=None):
+    def __new__(self, configfile = None, access_key=None, secret_key=None, access_token=None, profile=None):
         if self._instance is None:
             self._instance = object.__new__(self)
         return self._instance
 
-    def __init__(self, configfile = None, access_key=None, secret_key=None, access_token=None):
+    def __init__(self, configfile = None, access_key=None, secret_key=None, access_token=None, profile=None):
         if configfile:
+            debug("Config: Initializing with config file '%s', profile '%s'" % (configfile, profile))
             try:
-                self.read_config_file(configfile)
+                self.read_config_file(configfile, profile)
             except IOError:
                 if 'AWS_SHARED_CREDENTIALS_FILE' in os.environ or 'AWS_CREDENTIAL_FILE' in os.environ or 'AWS_PROFILE' in os.environ:
-                    self.aws_credential_file()
+                    self.aws_credential_file(profile)
 
             # override these if passed on the command-line
             # Allow blank secret_key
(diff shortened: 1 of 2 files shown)


Algirdas-Z (approved):

deividasraila:
@fviard please check

S3/Config.py
```diff
@@ -655,20 +677,29 @@ def parse_file(self, file, sections = []):
...
         with io.open(file, "r", encoding=self.get('encoding', 'UTF-8')) as fp:
             for line in fp:
                 if r_comment.match(line) or r_empty.match(line):
                     continue
                 is_section = r_section.match(line)
                 if is_section:
-                    section = is_section.groups()[0]
-                    in_our_section = (section in sections) or (len(sections) == 0)
+                    current_section = is_section.groups()[0]
+                    in_our_section = (current_section in sections) or (len(sections) == 0)
```
fviard: I think that there is legacy code, but except if I'm missing something, you can probably do directly:
`or not sections:` 
No need for the len.

S3/Config.py
```diff
@@ -261,18 +261,19 @@ class Config(object):
...
-    def __new__(self, configfile = None, access_key=None, secret_key=None, access_token=None):
+    def __new__(self, configfile = None, access_key=None, secret_key=None, access_token=None, profile=None):
         if self._instance is None:
             self._instance = object.__new__(self)
         return self._instance
 
-    def __init__(self, configfile = None, access_key=None, secret_key=None, access_token=None):
+    def __init__(self, configfile = None, access_key=None, secret_key=None, access_token=None, profile=None):
         if configfile:
+            debug("Config: Initializing with config file '%s', profile '%s'" % (configfile, profile))
```
fviard, line 271: Except this one, you can probably get ride of the other debug logs that I don't think are necessary once you will have a working code.

fviard (changes requested):
Thank you very much for your PR that is quite good and would solve an issue that is encountered by quite a few users and sorry for the long delay for my feedback.

If it would still be possible for you, there is a little change of behavior for this feature that I would prefer:
- Currently, for the s3cfg file, it will read one section or another (so either the default or the profile one).
- I think that it would be better to always try to load the "default" section, and to then optionally load the "profile" on top of that.
- It would probably be easy to do a first version with a slight modification of your code.
Something like creating the ConfigParser with the "[default]" sections, and then if there is the profile, use the "parse_file(...)" function of this cp instance with the "profile" as [sections].

I think that it would be better because if you have to copy paste all the common options in all your profiles, there is not that much any interest compared to use multiple config files. But with this, you could have a common (default) config and just put the specific parts inside each profile.

I added 2 nitpicks directly in review comments.

S3/Config.py
deividasraila: You are right, fixed.

S3/Config.py
deividasraila: Agree, removed all other newly created debug lines.

deividasraila:
@fviard I have updated and tested the code to always load default.

S3/Config.py
```diff
@@ -482,7 +483,8 @@ def aws_credential_file(self):
                     "Error reading aws_credential_file "
                     "(%s): %s" % (aws_credential_file, str(exc)))
 
-            profile = base_unicodise(os.environ.get('AWS_PROFILE', "default"))
+            if profile is None:
```
snosratiershad, line 486: As empty string can't be the file name case. how about `if profile:`?

S3/Config.py
```diff
@@ -559,8 +561,14 @@ def option_list(self):
             retval.append(option)
         return retval
 
-    def read_config_file(self, configfile):
-        cp = ConfigParser(configfile)
+    def read_config_file(self, configfile, profile=None):
+        if profile is None:
```
snosratiershad, line 565: Also here.

PR #1421, opened 2025-04-05
Add listacl listpolicy

marsteg:
I added some very simple commands to list the Policy and ACL. It seems like it's almost a design decission that this wasn't there... but since I missed it, i thought maybe others find it useful, too. 

Example Usage:
python3 s3cmd -c ceph.cfg listacl s3://test3
python3 s3cmd -c ceph.cfg listacl s3://test3/test.txt
python3 s3cmd -c ceph.cfg listpolicy s3://test3

I was only able to test this on ceph unfortunately.

### DIFF
--- a/s3cmd
+++ b/s3cmd
@@ -2174,11 +2174,11 @@ def cmd_sync(args):
 def cmd_getacl(args):
     cfg = Config()
     s3 = S3(cfg)
-    if len(args) != 1:
-        raise ParameterError("Too few parameters! Expected: <bucket>")
-    if not S3Uri(args[0]).has_bucket():
-        raise ParameterError("Invalid bucket name: '%s'" % args[0])
     uri = S3Uri(args[0])
+    
+    if uri.type != "s3" or not uri.has_bucket():
+        raise ParameterError("Expecting S3 URI instead of '%s'" % args[0])
+    
     try:
         acl = s3.get_acl(uri)
         acl_grant_list = acl.getGrantList()
--- a/s3cmd
+++ b/s3cmd
@@ -2171,7 +2171,7 @@ def cmd_sync(args):
         return cmd_sync_remote2remote(args)
     raise ParameterError("Invalid source/destination: '%s'" % "' '".join(args))
 
-def cmd_listacl(args):
+def cmd_getacl(args):
     cfg = Config()
     s3 = S3(cfg)
     if len(args) != 1:
@@ -2179,7 +2179,25 @@ def cmd_listacl(args):
     if not S3Uri(args[0]).has_bucket():
         raise ParameterError("Invalid bucket name: '%s'" % args[0])
     uri = S3Uri(args[0])
-    acl = list_acl(s3, uri)
+    try:
+        acl = s3.get_acl(uri)
+        acl_grant_list = acl.getGrantList()
+        owner = acl.getOwner()
+        output(u"Bucket:     %s" % uri.bucket())
+        output(u"Owner:      %s (ID: %s)" % (owner['nick'], owner['id']))
+        for grant in acl_grant_list:
+            output(u"   ACL:       %s: %s" % (grant['grantee'], grant['permission']))
+        if acl.isAnonRead():
+            output(u"   URL:       %s" % uri.public_url())
+        output(u"\nRaw ACL Output: \n %s" % acl)
+    except S3Error as exc:
+        # Ignore the exception and don't fail the info
+        # if the server doesn't support setting ACLs
+        if exc.status not in [404, 501]:
+            raise exc
+        else:
+            output(u"   ACL:       none")
+
     return acl
         
 def cmd_setacl(args):
@@ -2369,12 +2387,9 @@ def cmd_listpolicy(args):
     if uri.has_object():
         raise ParameterError("No Object Names allowed: '%s'" % args[0])
 
-    response = s3.list_policy(uri)
+    response = s3.get_policy(uri)
 
-    debug(u"response - %s" % response['status'])
-    if response['status'] == 200:
-        output(u"%s: Policy listed" % uri)
-    return EX_OK
+    return response
 
 def cmd_delpolicy(args):
     cfg = Config()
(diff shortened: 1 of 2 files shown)


mludvig (approved):
LGTM

fviard:
Thanks for your contribution, and sorry for the delay of my reply.
There is no dedicated command for list acl and list policy because today you will get both these info from the `s3cmd info` command.

I don't know if you have a case where you would really need a dedicated command for them?

For your info, there was 2 reasons for why there was not a dedicated command so far:
- s3cmd have already a lot of command, that makes it easily confusing, so avoiding adding much more than necessary is always better I think
- Especially for policy, a big question for adding a dedicated command is what format should be used for the output? Something formatted that could be convenient to understand by the user; the xml in output directly; an export to a file that could be used directly by the setpolicy, ...

fviard (changes requested):
I didn't put individual review comments as I don't think that the PR would be necessary as we have the "info" command with that,
but in the case that you would have a good argument for the dedicated commands, here are the issues I see with your PR:
- "list_policy" is not needed because "get_policy" already does that.
- Only bucket have policies, not the individual objects
- Your 2 new commands only output the input uri and nothing else...
(In fact, I discovered today that if you return a string to sys.exit, it will print it, so I see how you saw your acl, but we are not supposed to do that in the code, just return an error code at the end of the function)

marsteg:
I've updated the PR. To me, the explicit commands were useful. I cannot say how it is for others, but i think it's more confusing to expect the ACLs in an "info" command rather than having explicit one (i didn't find the info command 6 months ago, when I used this). But obviously this is your call :)

s3cmd
```diff
@@ -2346,6 +2375,22 @@ def cmd_setpolicy(args):
...
+    if len(args) != 1:
+        raise ParameterError("Too few parameters! Expected: <bucket>")
+    if not uri.has_bucket():
+        raise ParameterError("Invalid bucket name: '%s'" % args[0])
+    if uri.has_object():
+        raise ParameterError("No Object Names allowed: '%s'" % args[0])
+
+    response = s3.get_policy(uri)
+
+    return response
```
fviard: You should not return some content.
Return will be a "sys.exit" exit code indicating the result of the operation.
As here it will be a success, you should return:
`return EX_OK`

s3cmd
```diff
@@ -2346,6 +2375,22 @@ def cmd_setpolicy(args):
         output(u"%s: Policy updated" % uri)
     return EX_OK
 
+def cmd_listpolicy(args):
+    cfg = Config()
+    s3 = S3(cfg)
+    uri = S3Uri(args[0])
+
+    if len(args) != 1:
```
fviard: Once you will fix the argc value in the command list, you don't need this test anymore as it is aleady done before starting the command.

s3cmd
```diff
@@ -2171,6 +2171,35 @@ def cmd_sync(args):
...
+        raise ParameterError("Too few parameters! Expected: <bucket>")
+    if not S3Uri(args[0]).has_bucket():
+        raise ParameterError("Invalid bucket name: '%s'" % args[0])
+    uri = S3Uri(args[0])
+    try:
+        acl = s3.get_acl(uri)
+        acl_grant_list = acl.getGrantList()
+        owner = acl.getOwner()
+        output(u"Bucket:     %s" % uri.bucket())
+        output(u"Owner:      %s (ID: %s)" % (owner['nick'], owner['id']))
```
fviard: Good idea to show the owner, but you will have to do a special for "nick".
Based on what is on the following page, it will not be reported anymore by aws:
https://docs.aws.amazon.com/AmazonS3/latest/API/API_Grantee.html

I already don't see it.
So you could do something to only display it when it is available, but I think that it will be simpler and safer just to remove it and just display the id.

fviard: Owner field should have the same indentation as ACL.

fviard (changes requested):
I'm not 100% convinced that the dedicated commands are needed, but I'm not against adding them.
But there are a few more changes to be done that you will find with this review.

s3cmd
marsteg: i've basically completely copied the policy output from cmd_list now

s3cmd
```diff
@@ -2171,6 +2171,35 @@ def cmd_sync(args):
...
+        for grant in acl_grant_list:
+            output(u"   ACL:       %s: %s" % (grant['grantee'], grant['permission']))
+        if acl.isAnonRead():
+            output(u"   URL:       %s" % uri.public_url())
+        output(u"\nRaw ACL Output: \n %s" % acl)
+    except S3Error as exc:
+        # Ignore the exception and don't fail the info
+        # if the server doesn't support setting ACLs
+        if exc.status not in [404, 501]:
```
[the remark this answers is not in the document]
marsteg: I hope i understood you correctly. I just removed the if statement and output acl: none

s3cmd
```diff
@@ -2171,6 +2171,35 @@ def cmd_sync(args):
...
+        acl = s3.get_acl(uri)
+        acl_grant_list = acl.getGrantList()
+        owner = acl.getOwner()
+        output(u"Bucket:     %s" % uri.bucket())
+        output(u"Owner:      %s (ID: %s)" % (owner['nick'], owner['id']))
+        for grant in acl_grant_list:
+            output(u"   ACL:       %s: %s" % (grant['grantee'], grant['permission']))
+        if acl.isAnonRead():
+            output(u"   URL:       %s" % uri.public_url())
+        output(u"\nRaw ACL Output: \n %s" % acl)
```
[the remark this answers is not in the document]
marsteg: I was thinking it might be useful for non-trusting developers (like many of are :D ) to have a parameter, that would still show the Raw output. But I didn't understand how to add an optional parameter, so I removed it :(

s3cmd
```diff
@@ -2171,6 +2171,35 @@ def cmd_sync(args):
...
+    if len(args) != 1:
+        raise ParameterError("Too few parameters! Expected: <bucket>")
+    if not S3Uri(args[0]).has_bucket():
+        raise ParameterError("Invalid bucket name: '%s'" % args[0])
+    uri = S3Uri(args[0])
+    try:
+        acl = s3.get_acl(uri)
+        acl_grant_list = acl.getGrantList()
+        owner = acl.getOwner()
+        output(u"Bucket:     %s" % uri.bucket())
```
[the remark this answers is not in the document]
marsteg: copied bucket displaying from cmd_info

s3cmd
```diff
@@ -2171,6 +2171,33 @@ def cmd_sync(args):
...
+    
+    if uri.type != "s3" or not uri.has_bucket():
+        raise ParameterError("Expecting S3 URI instead of '%s'" % args[0])
+    
+    try:
+        acl = s3.get_acl(uri)
+        acl_grant_list = acl.getGrantList()
+        owner = acl.getOwner()
+        output(u"%s (bucket):" % uri.uri())
+        if owner['nick'] is None or owner['nick'] == "":
```
fviard: You can simplify a little bit by doing something like:
`if not owner['nick']:`
it will cover both cases.

fviard (commented):
Sorry for my long delay to reply sometimes and thank you for your own reactivity.
It is almost good, but it just remains a few small things and then we will be good to go :-)

btw I tested that it works ok for the nick with aws. In fact, for the moment, we will have the nick of the owner just for the bucket and not for buckets, so your code is working well in both cases.

fviard:
All good now and merged, thank you!

PR #1429, opened 2025-09-11
fix: cmd_object_put check not exists bucket

ChenMiaoi:
In the cmd_object_put function, after checking the type of the URL, verify whether the passed bucket exists.

The reason for doing this is that in a self-built S3 Server, I attempted to upload to a non-existent bucket using s3cmd, similar to 's3cmd put hello s3://b02', where b02does not exist, but the operation still succeeded(and the file could even be downloaded). This issue arises due to a design flaw in our S3 Server.

``` bash
$ ./s3cmd put setup.cfg s3://b02
upload: 'setup.cfg' -> 's3://b02/setup.cfg'  [1 of 1]
56 of 56   100% in    0s    11.63 KB/s  done
```

``` bash
$ mc put setup.cfg myio/b02
mc: <ERROR> unable to upload. Bucket `b02` does not exist.
```

However, similar clients like the MinIO client would directly return an error stating that the bucket b02 does not exist.Therefore, I believe it is necessary for s3cmd to include this check.

``` bash
$ ./s3cmd put setup.cfg s3://b02
ERROR: Bucket 'b02' does not exist
```

### DIFF
--- a/s3cmd
+++ b/s3cmd
@@ -402,6 +402,13 @@ def cmd_object_put(args):
         raise ParameterError("Destination must be S3Uri. Got: %s" % destination_base_uri)
     destination_base = destination_base_uri.uri()
 
+    check_uri = S3UriS3(destination_base)
+    check_bucket = check_uri.bucket()
+
+    if not any(bucket['Name'] == check_bucket for bucket in s3.list_all_buckets()['list']):
+        error(u"Bucket '%s' does not exist" % check_bucket)
+        return EX_NOTFOUND
+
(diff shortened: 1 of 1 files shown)


ChenMiaoi:
@fviard could u do the review, thanks

snosratiershad:
@mludvig @ChenMiaoi 
Yes, it's pretty much like checking a file's parent directory every time you want to write in. Also see [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use)

